PR #3, opened 2019-03-02
Add support for more Infohud queries

UNHchabo:
Adding:
MB's hp (different slot than normal boss HP)
Dash Counter
Vertical speed
Iframe timer

### DIFF
--- a/infohud.asm
+++ b/infohud.asm
@@ -90,7 +90,11 @@ ih_init_code:
 	STA $7FFB2C
 	STA $7FFB00
 	STA $7FFB30
-	STA $7FFB40
+	STA $7FFB38	; dash counter
+	STA $7FFB3A	; iframe counter
+	STA $7FFB3C	; vertical speed
+	STA $7FFB3E	; mother brain HP
+	STA $7FFB40	; enemy HP
 	STA $7FFB42
 	STA $7FFB44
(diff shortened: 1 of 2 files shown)


infohud.asm
```diff
@@ -567,6 +583,39 @@ charge_done:
 +	JMP Etanks	
 	;--/xfactor timer--
 
+	;--dash counter-- ;
+	DashCounter:
+	CLC
+	LDA $0B3F : AND #$00FF : CMP $7FFB38 : BEQ + : STA $7FFB38 : JSR Hex2Dec : LDX #$008A : JSR Draw3
+	LDX #$008A : JSR CheckLeadZero : LDA $7EC68E : CMP #$0057 : BNE + : LDA #$0C09 : STA $7EC68E
```
UNHchabo: At the moment this line is causing issues. On one build of the rom it made the shinespark counter never expire, where on another it made it crash. This is even without the Dash Counter display being used.

If we comment out this line it seems to work fine, just with leading zeroes on the dash counter.

Total suggesting switching to asar instead of xkas, because xkas could be generating broken code here.

UNHchabo:
I've commented out the CheckLeadZero line, and this seems to work perfectly for me, except the Dash Counter has leading zeroes.

PR #5, opened 2019-03-06
Fix build script and menu gfx

InsaneFirebat:
Fixed the path to xkas in the build script. Also added the build folder since the script won't output anything if it doesn't exist. Contains an empty file because GitHub doesn't seem to allow an empty folder.

Sorry if this looks confusing. Still learning how GitHub works.


InsaneFirebat:
Added in a fix for the broken menu graphics.

The original project had these graphics changes pre-patched into the rom instead of the source.

PR #11, opened 2020-01-06
Add support for GT Classic presets

InsaneFirebat:
GT Classic presets added. Hi Jump and Bootless routes combined into one category.

Bumped version to 2.0.8


InsaneFirebat:
Are the failed Azure checks normal? I've tested the changes in the rom with no issues but I can only assume I copy/pasted correctly on the website stuff :)

Makefile
```diff
@@ -16,4 +19,7 @@ sd2snesrbo:
 sd2sneskpdr25:
 	cp resources/sm_orig.sfc build/smhack20_sd2snes_kpdr25.sfc && cd src && asar --no-title-check -DFEATURE_SD2SNES=1 -DCATEGORY=2 main.asm ../build/smhack20_sd2snes_kpdr25.sfc && cd -
 
-all: sd2snes smhack sd2snesrbo smhackrbo sd2sneskpdr25 smhackkpdr25
+sd2snesgtclassic:
+	cp resources/sm_orig.sfc build/smhack20_sd2snes_gtclassic.sfc && cd src && asar --no-title-check -DFEATURE_SD2SNES=1 -DCATEGORY=3 main.asm ../build/smhack20_sd2snes_gtclassic.sfc && cd -
+
+all: sd2snes smhack sd2snesrbo smhackrbo sd2sneskpdr25 smhackkpdr25 sd2snesgtclassic shmackgtclassic
```
UNHchabo: "shmackgtclassic" should be "smhackgtclassic".

InsaneFirebat: Good catch. Resolved now.

azure-pipelines.yml
```diff
@@ -1,71 +1,73 @@
-# ASP.NET Core
```
UNHchabo: Can you check if the newlines got changed for this file or something? It would help to just have the changed lines in the diff.

InsaneFirebat: I honestly have no idea what causes this, or how to check newline usage. When I look at the diff in GitHub's desktop client, it only shows lines 54-55 having differences. WinMerge, my usual diff tool, also points to lines 54-55.

InsaneFirebat: My initial commit shows a proper diff, but the lowercase g fix seems to have messed it up. I may have committed this change directly from the browser.

First link shows the two lines added, second link shows all text as being replaced but also highlights the actual fix on those new lines.
https://github.com/InsaneFirebat/sm_practice_hack/commit/c16e6daa20a51d0d3c0566456d6af686d991fde6#diff-fec826feae04e51c0d94076385408bdc
https://github.com/InsaneFirebat/sm_practice_hack/commit/bf04e4f12683fbae25a6666d9f18f0b169e28934#diff-fec826feae04e51c0d94076385408bdc

InsaneFirebat:
That last commit fixed the diff. I'll know better than to use the browser for editing in the future.

PR #15, opened 2020-01-15
Add KPDR 21% to combined ROM

InsaneFirebat:
This adds the KPDR 21% presets to the combined ROM.


UNHchabo:
Did you want to also remove the KPDR21 rom?

InsaneFirebat:
KPDR 21% option removed.

PR #19, opened 2020-05-29
New presets for 14% Speed and All Bosses

InsaneFirebat:
14ice rom category renamed to "14low"
14% Speed presets added to 14low category
Three Any% All Bosses routes added (KPDR, PKDR, PRKD)
README updated with known issues

PR #20, opened 2020-05-31
New presets for 14% Speed and All Bosses (InsaneFirebat)

tewtal:
This is PR #19 rebased onto master with the menu code rewritten to handle menu/preset data split into separate banks so that we can combine all the presets into a single ROM.

Original PR comments below:
14ice rom category renamed to "14low"
14% Speed presets added to 14low category
Three Any% All Bosses routes added (KPDR, PKDR, PRKD)
README updated with known issues


tewtal:
@InsaneFirebat I've reworked your latest PR into this one and combined it with the latest changes I've made to master. If you can I'd be happy if you could take a quick look at the code and see if I screwed something up in the process before I merge this.

InsaneFirebat:
The preset stuff looks good, but I'm not getting a modified rom out of the build script. I just get sm_orig.sfc in the build folder, and it's vanilla.

tewtal:
It's a change with the new build system, it now generates IPS patches directly instead of patching a ROM. I changed it to do this so that I don't have to include a original ROM into the build process for the Docker container. You should have two IPS-files in the build/ folder that you can apply to an SM ROM.

tewtal:
Another option is to just place an original ROM-file with the same name as the IPS in the build folder and some emulators will automatically patch the game with the IPS file when it loads the ROM. I know Mesen-S does at least which I've done development with.

InsaneFirebat:
I see sm_orig.sfc and 00.sfc being created, then ff.sfc, then 00 and ff are deleted. I'm not getting any IPS files in the build folder. Is the process listed in README still valid? Vanilla ROM in resources and asar stuff in tools?

tewtal:
The windows build.bat was broken, I've fixed that in master and rebased this PR on top of that change. One change is that you now will need python 3 installed as this is used to create the patches. You don't need to have any vanilla ROM in resources any more. I'll have to update the build instructions as well I see :)

InsaneFirebat:
Wow, it doesn't even need need a rom to compare to? That's cool.

I've got python installed now and tested the roms. All seems good on the first pass. You may want to mention the PATH environmental variable if you're editing the README.

tewtal:
I'll go ahead and merge this for now then so it gets pushed to the website and hopefully nothing breaks :) If it does we'll have to just fix it

PR #24, opened 2020-07-19
Added Shine Tune and Shot Timer to the Infohud Mode options

idlechild:
Hi!
Please review before merging.  While I have plenty of C/C++ experience, this is more work in assembly than I have ever done.  Fortunately your source code was easy enough to follow, and the SM RAM maps available out there helped fill in the gaps.  The changes looked good in testing.

The shot timer is easy enough to understand (time between initially pressing the shoot button, hopefully useful for learning doppler).  For the "shine tune" feature, I made a video.  I plan to redo the video because of the bad audio/video sync (sorry!) and also I made an incorrect statement in there and I forgot to talk about stuttering.  Despite all that, it's good enough that you can see what the shine tune does.  I start demoing shine tune halfway through the video:
https://www.twitch.tv/videos/684151396

Edit: also forgot to mention you mention cases where shine tune numbers will be overwritten, for example when transitioning to another room.  Also I'll drop the complaint about the input viewer; is that just me?  I find it's very sluggish when using the original practice hack version 2.0.15 or my 2.1.1 version, but it works fine with the vanilla rom.  Then today I see Zeni using the practice hack with input viewer and his is working fine.

Edit: Made a better video, included stuttering this time
https://www.twitch.tv/videos/684595233

### DIFF
--- a/src/defines.asm
+++ b/src/defines.asm
@@ -6,12 +6,15 @@
 !ram_transition_counter = $7FFB0E
 !ram_last_door_lag_frames = $7FFB10
 
+!ram_armed_shine_duration = $7FFB14
 !ram_etanks = $7FFB12 ; ??
 !ram_max_etanks = $7FFB24 ; ??
 !ram_last_hp = $7FFB9A
 
 !ram_slowdown_mode = $7EFFFC
 !ram_slowdown_frames = $7FFB52
+!ram_shine_dash_held_late = $7FFB1A
+!ram_shine_counter = $7FFB30
 !ram_cooldown_counter = $7FFB32
 !ram_xpos = $7FFB34
 !ram_ypos = $7FFB36
@@ -20,9 +23,6 @@
 !ram_vertical_speed = $7FFB3C
 !ram_mb_hp = $7FFB3E
 !ram_enemy_hp = $7FFB40
-!ram_shine_counter_1 = $7FFB30 ; goes 1-A
-!ram_shine_counter_2 = $7FFB14 ; armed shine duration
-!ram_shine_counter_3 = $7FFB1A ; armed shine duration 2
 !ram_magic_pants_1 = $7FFB64
 !ram_magic_pants_2 = $7FFB66
 !ram_magic_pants_3 = $7FFB70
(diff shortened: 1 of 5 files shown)


src/defines.asm
```diff
@@ -20,9 +23,6 @@
 !ram_vertical_speed = $7FFB3C
 !ram_mb_hp = $7FFB3E
 !ram_enemy_hp = $7FFB40
-!ram_shine_counter_1 = $7FFB30 ; goes 1-A
-!ram_shine_counter_2 = $7FFB14 ; armed shine duration
-!ram_shine_counter_3 = $7FFB1A ; armed shine duration 2
```
idlechild, line 25: Let me know if want more explanation of changes.

Here I got rid of shine_counter_1 because it goes 1-A like the comment said... and that's it... I can't find any other reason for it to exist.  I gave the other two variables better names.

Also one of the variables (shine_dash_held_late) is now used for shine tune feature.

src/infohud.asm
```diff
@@ -586,6 +581,136 @@ status_cooldowncounter:
     RTS
 }
 
+status_shinefinetune:
```
idlechild, line 584: This method is organized to reduce JMP statements.  I only ended up need one.  It's nice that the compiler tells us when the relative branch jumps are too long.

src/mainmenu.asm
```diff
@@ -818,6 +818,8 @@ ih_display_mode:
     db #$28, " X POSITION", #$FF
     db #$28, " Y POSITION", #$FF
     db #$28, "   COOLDOWN", #$FF
+    db #$28, " SHINE TUNE", #$FF
+    db #$28, " SHOT TIMER", #$FF
```
idlechild, line 822: I did find the code that allows the user to scroll left or right in the menu, and it already had support for 12 items in the list even though there were only 10, so this time I could just add items here.  Next time we'll have to update the menu scrolling logic.

InsaneFirebat:
Nice work. I really enjoyed playing with the shine tune feature. I've shared it with a few people and it seems to be well received by those that use the short charging tools. Thanks for making the videos as well.

I haven't encountered any issues in the last few days of practicing with these changes.

Issue #26, opened 2020-07-31
group related infoHUD options next to each other


krehera:
current order:

```
- Enemy HP

- Shinespark

- Charge

- X-factor

- MB HP

- Dash

- Vert Speed

- I-frames

- Lag Counter

- X-position

- Y-position

- cooldown

- shine tune
```

recommended order:
```
- Enemy HP

- MB HP

- Charge

- X-factor

- cooldown

- Shinespark

- Dash

- shine tune

- I-frames

- Lag Counter

- X-position

- Y-position

- Vert Speed
```

idlechild:
I am fine with the reorder if people don't mind the order changing every so often.  If we put related items together, then every time a new option is added and you previously saved with an option (say I-frames) then when you update the ROM it might change (in this case to shine tune).  It's not a big deal though.  I wonder if most people start a new save file every time, or if they just leave it on enemy HP and wouldn't notice the change.

I'm biased towards this reorder so I can put Room Strat as the second option someday.  I'm thinking there could be options that only work in a specific room, so instead of listing them as separate options, you would pick the Room Strat option and then choose which room strat in a second menu.  I didn't want Room Strat to be buried in the option menu though.  If MB HP becomes a room strat then that's convenient.

PR #28, opened 2020-09-03
Reorder infohud options, add jump counter, X/Y subpixels, moat CWJ room strat, improve shine tune

idlechild:
Infohud items have been reordered per issue https://github.com/tewtal/sm_practice_hack/issues/26.  The shot counter and newer jump press counter (number of frames jump was held) are at the end.

X and Y position now show subpixels in the second column (replacing HP).  Subpixels are rendered in hex (inverted 0-5 are used in place of A-F).

A new "room strat" menu option has been added which contains MB HP and Moat CWJ.  The following video or shorter clip are for the Moat CWJ feature.  The before-after idea was also incorporated into shine tune so there is more feedback when missing the short charge tap.
https://www.twitch.tv/videos/729670980
https://www.twitch.tv/idlechild/clip/OptimisticFurryChamoisAliens
(the last comment "you might see nothing" happens if you jump so early that the fanfare eats your jump input)

The before-after idea is to display B1 if you jumped (CWJ) or ended tap (shine tune) one frame before you should have, B2 for two frames before, etc. up to BX if 10 or more frames.  Same for A1, A2, ... AX.  Honestly I would have preferred E1, E2, ... for early and L1, L2, ... for late, but it looks like we did something special to render A,B,L,R,X,Y in the infohud and I wasn't sure how to render E.


InsaneFirebat:
> it looks like we did something special to render A,B,L,R,X,Y in the infohud and I wasn't sure how to render E.

The graphics for input display should be in resources/menugfx.bin. I'm not sure how it's formatted or if it's compressed. The original source release for v1.42 was missing these characters so I grabbed the diff from a hex editor and saved it as binary data to be incsrc'd in infohud.asm. This was discussed a bit in the #smhacking channel back in early March 2019. Total may have more insight if that doesn't solve your issue.

idlechild:
Thanks @InsaneFirebat, that was helpful.  I added 16 bytes to the menugfx.bin file.  It looks like the original ROM contents at that location were the plain X so I think it is okay to replace it with an E.

idlechild:
Okay all done with this pull request, I think it is ready.  Not sure if anyone wants to try it out first or just merge it.

tewtal:
I'll go ahead and merge it, worst case if something breaks we can just fix it or revert it. Thanks for the contribution!

PR #29, opened 2020-09-29
Update create_ips.py

awilkester:
per the previous fix submitted and found by chabo in "create_dummies.py": 

fo.write(bytes) needs to be fo.write(bytearray), otherwise strings are written to the file, not the binary data

image


UNHchabo:
Good catch, thanks.

PR #32, opened 2021-01-31
Added Botwoon CF Room Strat and updated Spikesuit

idlechild:
Trainer for learning the crystal flash clip in botwoon hallway.  Also spikesuit trainer now checks if springball is active, and if so then doesn't count jump as unmorph.

I can update this to version 2.1.5 if we accept IFB's pull request first, or perhaps we want to take them both at the same time?


idlechild:
Finished my Botwoon CF testing with bombs, springball and suitless
Here's springball:
https://clips.twitch.tv/GiantTemperedGuanacoImGlitch

Also just got a request from Stashiocat to show shine tune numbers through door transition, so now hiding the segment timer when shine tune is selected.

IMO this pull request is ready.

Issue #12, opened 2020-01-06
Crash before endgame cutscenes


InsaneFirebat:
The game will crash after the ship leaves exploding Zebes and the screen fades to black. Just set the "Zebes Set Ablaze" event flag and enter the ship to quickly reproduce.

This also occurred in v1.31 of InfoHUD and was fixed by v1.41, so Total may have some insight.

InsaneFirebat:
PR #39 fixes this issue.

PR #38, opened 2021-04-25
Submenus, PKRD Presets, and MB Splits

InsaneFirebat:
This pull adds submenus for selecting Category Presets, InfoHUD Display Modes, and Room Strats. These lists have grown a bit too large for the left/right selection. It's much easier to see the choices and find what you want now. The old menus have been repurposed to show which option is currently selected, but can still be used the old way. Preset Category was moved out of Misc and into the new submenu.

The segment splits for Mother Brain occur when control is returned to Samus (just before music change) in the MB1 cutscene, and again when Baby is spawned off-screen (just before final rainbow charge).

The option to Clear Shortcuts was added because I've seen a lot of runners set things by mistake when first checking out the practice rom. This will unset/zero all of the configured shortcuts except Main Menu, which will be reset to default.

PKRD Presets and doc updates included. Help page updates are "blind" edits, as usual.


InsaneFirebat:
I've added in a fix for Botwoon's RNG control which has been bugged since the 2.0 release. This was caused by the menu index being written to the wrong address as well as being decremented before bit shifting.

InsaneFirebat:
One more commit added to fix the endgame crash when leaving Zebes. This was caused by a decompression routine incrementing sram values and writing to !ram_load_preset. My fix is to skip loading presets if gamemode = 27 (credits).

Fixes Issue #12

PR #39, opened 2021-05-03
Submenu, PKRD, MB Segments, and Fixes

InsaneFirebat:
Added submenus for selecting preset categories and HUD modes. Recycled old left/right menu as a current selection indicator.
Added presets for Any% PKRD.
Update HUD between Mother Brain phases. Share code with UNHchabo's elevator update.
Added an option to clear controller shortcuts while resetting the main menu shortcut to default.
Fixed the crash at endgame cutscenes by checking gamemode before loading a preset.
Fixed Botwoon RNG control with correct address and bit manipulation.


src/mainmenu.asm
```diff
@@ -905,7 +1108,7 @@ rng_rerandomize:
 
 rng_botwoon_rng:
     dw !ACTION_CHOICE
-    dl #$7FFB8A
+    dl #!ram_botwoon_rng
```
InsaneFirebat, line 1111: $7FFB8A was the address used in the original implementation for v1.42. It was moved to $7FFB86 in defines.asm. This was only half the problem.

The old menu (v1.42) wrote specific values for each choice, while the current menu can only pass the menu index. hook_botwoon_rng needs to convert the index to those specific values, or branch out if zero (set to random). More on that in the rng.asm comment.

src/rng.asm
```diff
@@ -147,7 +147,7 @@ hook_botwoon_rng:
 
     LDA !ram_botwoon_rng : BEQ .no_manip
 
-    DEC : ASL #3
+    DEC : ASL #3 : INC
```
InsaneFirebat, line 150: Link to old menu source with needed values for RNG manip: https://github.com/tewtal/sm_practice_hack/blob/12cc9c87c421c8a39720a74ed210c1444f424787/menu.asm#L1066-L1071

`DEC : ASL : ASL : ASL : INC`
8 * (X - 1) + 1 = !ram_botwoon_rng
Random = 0
Down = 1 -> 1
Up = 2 -> 9
Right = 3 -> 17
Left = 4 -> 25

src/mainmenu.asm
```diff
@@ -835,11 +1000,49 @@ ih_display_mode:
...
+
+ihstrat_elevatorcf:
+    %cm_jsr("Elevator Crystal Flash", #action_select_room_strat, #$0004)
+
+ihstrat_robotflush:
+    %cm_jsr("Robot Flush", #action_select_room_strat, #$0005)
+
+action_select_room_strat:
+{
+    TYA : STA !sram_room_strat
```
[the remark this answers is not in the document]
idlechild: Good idea

idlechild:
Just noticed you raised this pull request.  This all looks great!  This is worthy of bumping the revision to 2.2.0 instead of 2.1.10.

InsaneFirebat:
I have Draygon RNG control working, but haven't had much time to test it. I'd like to get that in before jumping to 2.2. Should I pull that into this request or give it a PR of its own?

It's here in my fork.
https://github.com/InsaneFirebat/sm_practice_hack/commit/c9e0e7ba99a7ff8bdedf6c5bbee6ff8719d7c2f5

InsaneFirebat:
Draygon control is in now as well as the Room Strat menu change mentioned above. I moved Magic Pants from InfoHUD to Misc because it makes more sense there to me.

idlechild:
Draygon RNG poggers!

I'm on the fence on moving magic pants.  It makes a bit more sense where you put it, maybe that's worth the occasional "what happened to magic pants?" post on discord.

Let us know if you're waiting on anything before calling it 2.2 and merging it, whatever you decide on the magic pants looks good to me.

InsaneFirebat:
This should be ready for merging now.

InsaneFirebat:
I misclicked and submitted that before adding in the docs. Crocomire RNG is now part of the pull. You can choose between the step forward attack, or the swipe and shoot attacks.

InsaneFirebat:
Added RNG Control option to set Phantoon's eye close delay. Still ready for merge on my end.

InsaneFirebat:
Latest commit expands the Draw4Hex subroutine to be able to draw hex values with A-F or the current 0-5 with inverted colors. There's an option in the menu to choose your preferred style, defaulted to the inverted 0-5. I added some more tiles to menugfx.bin and renamed it to hudgfx.bin. I wanted to have my hex tiles all on the same line so I ended up filling a few blanks with the characters N, decimal, and hyphen. The decimal is being used for X/Y pos/speed with the other two currently unused.

InsaneFirebat:
Pattern and direction can now be set independently for Phantoon. New event toggles for metroid rooms and chozo acid.

src/mainmenu.asm
```diff
@@ -746,13 +746,27 @@ EventsMenu:
     dw #events_goto_bosses
     dw #events_zebesawake
     dw #events_maridiatubebroken
+    dw #events_chozoacid
```
idlechild, line 749: I like this event, but not sure what the metroid room events do, those might just be wasting space

src/defines.asm
```diff
@@ -110,6 +110,7 @@
 !sram_save_has_set_rng = $70201E
 !sram_preset_category = $702020
 !sram_room_strat = $702022
+!sram_hexstyle = $702028
```
idlechild, line 115: You skipped a couple slots here

src/infohud.asm
```diff
@@ -1896,6 +1924,9 @@ NumberGFXTable:
     dw #$0C09, #$0C00, #$0C01, #$0C02, #$0C03, #$0C04, #$0C05, #$0C06, #$0C07, #$0C08
     dw #$0C45, #$0C3C, #$0C3D, #$0C3E, #$0C3F, #$0C40, #$0C41, #$0C42, #$0C43, #$0C44
 
+HexGFXTable:
```
idlechild, line 1927: This is a bad spot for this table, it will affect number overflow.  Try fighting MB3 with MB HP option and you'll see why.

Either move this to the top or to the bottom of existing tables.

src/infohud.asm
```diff
@@ -436,6 +440,35 @@ ih_update_hud_code:
     RTL
```
idlechild, line 440: A few lines up from here, there is existing code that prints a decimal.  Just saying you don't need a new graphic.

src/infohud.asm
```diff
@@ -436,6 +440,35 @@ ih_update_hud_code:
...
+
+    ; Run standard code and return
+    PLY
+    PLX
+    PLA
+}
+
 ih_hud_code:
 {
     %ai16()
```
idlechild, line 474: About 50 lines down, there is code to print hp.  Can you add a line to write a space over the decimal at 7EC690?

idlechild:
I like most of the changes, except for the rework phantoon rng control commit.  Of course I like the other changes such as eye close control.  However separating the direction and pattern just adds menuing, and there is another problem.  If you lock in the 2nd round direction but leave pattern random, the rng won't work the same.  You will get a lot of rare patterns.  So I don't see the point of separating direction and pattern; I think we should revert that.

Sorry to be harsh!  Code reviews can be rough.

InsaneFirebat:
All good. That's probably more feedback than I've had on all my other pulls combined, so I see it as a positive :)

The thing with rare Phantoon patterns was partially known. However, I didn't think to check for that with both set to random. I don't mind it when forcing a direction (kinda like it tbh), but I don't want to affect RNG with everything set to random. I'll definitely revert that.

src/infohud.asm
InsaneFirebat: Fair enough. I really just needed something to put there, and wanted it to be something useful instead of blank/garbage tiles. I'll swap to the old decimal for these in case the new one gets replaced later.

src/defines.asm
InsaneFirebat: I have some settings in custom builds that use 24 and 26. I'll move some stuff around to fix that.

I'd like to go through that entire file and reorganize, but I don't want to mix that in with other changes.

src/mainmenu.asm
InsaneFirebat: Once you've killed all of the metroids in a room, they will no longer spawn. Only way to get them back is to load a preset (which would change your ammo and other events) or reset all events.

I have the Zebetite flags in my custom builds, but I left those out of the pull because I haven't figured out how the 4th zeb works.

src/infohud.asm
```diff
@@ -521,6 +521,7 @@ ih_hud_code:
     JSR (.status_display_table,X)
 
     ; Samus' HP
+    LDA #$0C0E : STA $7EC690 ; erase stale decimal tile
```
idlechild: I think it's okay to do this after the comparison; you might see the decimal after changing categories but only until HP changes.  I think we should avoid affecting lag wherever possible.

idlechild: Wait you already changed it.  What?  How are you reading my mind???

src/infohud.asm
idlechild: Or just make everything awesome, that works too :)

idlechild (approved):

src/infohud.asm
InsaneFirebat: It was blocking my % for CPU Usage :)

src/infohud.asm
InsaneFirebat: I thought the new submenus were going to be the star of this show, but the number are really making me happy. Thanks for pointing this out.

PR #42, opened 2021-06-05
When fanfare is off, music is not interrupted

idlechild:
Posting this for review now.  I think it works, but feel free to critique it as I'm not 100% confident on it.  Also let me know if you think there are side effects such as impact to timing or lag.

I don't plan to merge this in until I finish other changes; I'll update the changelog at that time.

### DIFF
--- a/src/fanfare.asm
+++ b/src/fanfare.asm
@@ -1,371 +1,78 @@
-; $84:8C01 22 C1 8F 80 JSL $808FC1[$80:8FC1]
-org $848C01
-    JSL hook_queue_music_track
+; $84:EFD7: Just past start of free space
+org $84EFD7
+  .prepareloop
+    PHX              ; start of logic that was overwritten
+    LDX #$000E
+
+  .clearloop
+    STZ $0619,x
+    STZ $0629,x
+    DEX
+    DEX
+    BPL .clearloop
+    BRL .continue    ; jump back to the original logic
 
 
-; $85:8493 20 36 81    JSR $8136  [$85:8136]  ;\
+; $84:8BDD: Instruction - clear music queue and queue music track [[Y]] ;;;
+org $848BDD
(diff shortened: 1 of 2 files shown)


InsaneFirebat:
On first inspection, I noticed that Ceres starts with no music. That's a feature, right? It plays if you toggle the music on/off, or advance to Ridley. Nice work.

src/infohud.asm
```diff
@@ -82,7 +82,7 @@ incbin ../resources/hudgfx.bin
 
 
 ; Main bank stuff
-org $DFE000
+org $E08000
```
idlechild, line 85: We aren't out of space yet, but still might as well give infohud its own bank

src/infohud.asm
```diff
@@ -323,10 +323,10 @@ ih_update_hud_code:
         }
 
         ; Draw seconds
-        LDA !ram_tmp_1 : JSR Hex2Dec : LDX #$003C : JSR Draw3
+        LDA !ram_tmp_1 : LDX #$003C : JSR Draw3
```
idlechild, line 326: I combined Hex2Dec with the Draw methods, which made the methods more efficient.  It also allows for better number overflow.  If a two-digit or three-digit number overflows, it will use the new 10-65 characters.

src/infohud.asm
```diff
@@ -561,35 +561,43 @@ status_roomstrat:
     RTS
 
   .status_room_table
-    dw status_mbhp
+    dw status_tacotank
+    dw status_gateglitch
```
idlechild, line 565: Gate Glitch added, Taco Tank is just a placeholder.
Once Taco Tank is done, I plan to merge in this pull request.

I'm also planning to make a new .asm file just for the infohud modes and room strats, since this file is a bit large.

PR #44, opened 2021-06-16
Update 2.2.3

InsaneFirebat:
**Automatically select Power Bombs when loading certain presets.** teledump.lua will now record the currently selected item in preset data. We should be careful not to equip missiles or supers when making new presets. Most existing categories have been manually updated to equip power bombs at certain presets where they will ALWAYS be used in the next room.

**Added RAM Watch HUD mode with a new InfoHUD menu.** RAM Watch can be used to monitor any address in $7E space. Addresses are configured in a separate submenu within the InfoHUD menu. Optionally, values can be written to the configured addresses or "locked" to be rewritten every frame (as long as RAM Watch is selected as the current HUD mode). This required adding an additional number field menu item that supports hexidecimal.

**Update HUD timers when boss drops spawn.** Requested feature. Runs ih_update_hud_early during boss drop spawn routines.

**Reset segment timer at game start for practice run timing.** Requested feature. Previously, the segment timer began running shortly after power on. It will now reset on START GAME and initialize to 1:50 to account for the frames it will miss as Ceres is loading. There was an existing hijack that reset all timers after completing Ceres. This has been changed so that it adds 1:13 to the segment timer to account for frames missed while loading Zebes and cutscenes. All other timers are still reset at this time. With these changes, the segment timer can be used to approximate RTA timing in a practice run or theory TAS. Note that starting from a fresh file will add 20 frames to the Ceres loading sequence. This is not accounted for.

I've also moved the sprite priority flag into SRAM so that it won't be reset when loading a preset or savestate.

### DIFF
--- a/src/infohudmodes.asm
+++ b/src/infohudmodes.asm
@@ -736,8 +736,16 @@ status_vspeed:
   .newjump
     LDA #$0001 : STA !ram_roomstrat_counter : STA !ram_walljump_counter
 
+    ; Print initial jump speed over item%
+    LDA $0B38 : BNE +
+    LDA $7EC612 : STA $14
+    LDA $0B2D : AND #$0FFF
+    LDX #$0012 : JSR Draw4Hex
+    INC $0B38
+    LDA $14 : STA $7EC612
+
     ; If we started falling and space jump might be allowed, time to compare
-    LDA !ram_roomstrat_state : BEQ .done
++   LDA !ram_roomstrat_state : BEQ .done
     BRL .preparecompare
 
   .resetjumpcounter
--- a/src/infohudmodes.asm
+++ b/src/infohudmodes.asm
@@ -737,12 +737,12 @@ status_vspeed:
     LDA #$0001 : STA !ram_roomstrat_counter : STA !ram_walljump_counter
 
     ; Print initial jump speed over item%
-    LDA $0B38 : BNE +
-    LDA $7EC612 : STA $14
+    LDA $0B1A : BNE +
+    LDA $7EC612 : TAY
     LDA $0B2D : AND #$0FFF
     LDX #$0012 : JSR Draw4Hex
-    INC $0B38
-    LDA $14 : STA $7EC612
+    INC $0B1A
+    TYA : STA $7EC612
 
     ; If we started falling and space jump might be allowed, time to compare
 +   LDA !ram_roomstrat_state : BEQ .done
--- a/src/infohudmodes.asm
+++ b/src/infohudmodes.asm
@@ -23,6 +23,7 @@
     dw status_quickdrop
     dw status_walljump
     dw status_shottimer
+    dw status_ramwatch
 
 status_enemyhp:
 {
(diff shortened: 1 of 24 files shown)


InsaneFirebat:
Added initial jump speed to the vertical speed HUD mode. It overwrites the item collection percentage with a three digit hex value that shows speed and subspeed when jumping from a solid tile.

src/infohudmodes.asm
```diff
@@ -735,8 +736,16 @@ status_vspeed:
   .newjump
     LDA #$0001 : STA !ram_roomstrat_counter : STA !ram_walljump_counter
 
+    ; Print initial jump speed over item%
```
idlechild, line 739: There's the max etanks number below item%.  Perhaps we could Draw3 the whole number speed (upper byte of 0B2D) over item% and then Draw2Hex the fractional speed (lower byte of 0B2D) over the max etanks.

src/infohudmodes.asm
```diff
@@ -735,8 +736,16 @@ status_vspeed:
...
+    LDA $0B38 : BNE +
```
idlechild, line 740: Is this check necessary?  Why not always print the speed if we are here? (jump just pressed)

src/infohudmodes.asm
```diff
@@ -735,8 +736,16 @@ status_vspeed:
...
+    LDA $7EC612 : STA $14
+    LDA $0B2D : AND #$0FFF
+    LDX #$0012 : JSR Draw4Hex
+    INC $0B38
```
idlechild, line 744: Not a fan of incrementing RAM that the game might use.  What if incrementing the value from 0 to 1 (before the game itself does it) messes with the physics in some way?

I guess I'm asking if we're sure this value is only written to by the game and not read from.

idlechild:
Overall looks good.  Seems like a lot of new variables, but I think they are all needed.  I'm just not sure about the initial jump logic.

src/infohudmodes.asm
InsaneFirebat: The whole number would always be a single digit afaik. I haven't seen it go over 9 yet. I'd do it if you think it would be easier for users to figure out.

src/infohudmodes.asm
InsaneFirebat: Wall jumps and space jumps have always predictable starting speeds. Normal jump height can be a bit unpredictable with speed booster equipped and dash held (with or without boots). I don't want the value to disappear if you have to wall jump or space jump after the jump that needs to be measured.

src/infohudmodes.asm
InsaneFirebat: I was testing both values, but meant to use $0B1A in the end. $0B2D does get read in a few unknown routines, but I'm confident that $0B1A is only zeroed in vanilla code. I poked through all of the banks in the disassembly with Ctrl+F to confirm, plus breakpoints on memory reads.

InsaneFirebat:
I have another commit sitting around with some extra preset adjustments. Let me know how you feel about the change from $0B38 to $0B1A and I'll push it out.

src/infohudmodes.asm
idlechild: Yeah that's the main reason I was suggesting it, just easier to understand

...although it might be more obnoxious to read... maybe it's fine the way you have it.

idlechild:
Sounds good, I'd be happier with $0B1A.

PR #45, opened 2021-06-27
Chozo grab segment and door stuck fix

InsaneFirebat:
HUD timers will now update when bowling and acid chozo statues grab onto Samus.

The door transition flag is now cleared when loading a preset. This was the cause of the "door stuck" bug after loading a preset from a door transition. Only rooms with elevators were affected. The "handle door transitions" routine at $82E17D checks if the elevator is active ($0E16) to know if it's handling a door. The elevator is active when Samus is standing still on it, but the main elevator routine ($A3952A) branches out immediately if the door transition flag is set. This makes the elevator unresponsive to Samus pressing down and unable to clear its own flag when Samus steps off. 

You can write a 1 to $0795 after loading a Brinstar Elevator preset to artificially reproduce the issue. When you go through a door, the game will assume Samus needs to be locked for the elevator ride.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,5 +15,4 @@ The build script will create two IPS patch files for the practice hack. The patc
 
 ## Known Issues:
 
-* Loading a preset while in a loading transition can sometimes cause Samus to freeze when exiting the next door.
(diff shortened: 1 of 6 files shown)


idlechild (commented):
Looks good to me.

I am hoping to add to the 2.2.4 update, getting close to making the minimap work.

PR #46, opened 2021-07-01
Added minimap option, debug CPU brightness option, Kraid claw RNG

idlechild:
This includes InsaneFirebat's 2.2.4 pull request

There's still a couple issues.  The infohud tilemap isn't loading right if you use presets.  Also if you have minimap enabled, certain actions (entering boss fight, entering pause menu, others?) cause the original practice rom tileset to be loaded when it should use the minimap tileset.  Opening/closing the practice rom menu fixes both issues, but of course would like to get these issues resolved before merging.

Update: These issues are fixed except for Kraid (Kraid has special VRAM logic).  I may add a clear map option, but otherwise this PR might be ready.


src/infohud.asm
```diff
@@ -312,29 +341,36 @@ ih_mb1_segment:
...
 }
 
 ih_drops_segment:
 {
     ; runs when boss drops spawn
     JSL ih_update_hud_early
-    JSL $808111 ; overwritten code
-    RTL
+    JML $808111 ; overwritten code
+}
```
InsaneFirebat, line 360: Looks like the JML landed in the wrong spot on this one.

src/minimap.asm
```diff
@@ -0,0 +1,214 @@
...
+    NOP : NOP : NOP
+
+org $82AEAF      ; routine to remove auto reserve icon on HUD from equip screen
+    NOP : NOP : NOP
+
+org $809AF3
+    JSL mm_initialize_minimap
+
+org $90A91B
+    LDA !ram_minimap : BNE .update_minimap
```
idlechild, line 18: Tracking CPU impact of the minimap logic (when minimap is off) that occurs every frame.  There's some impact to other areas such as ih_update_hud_code but that only runs at specific times.

This LDA : BNE adds 6 CPU cycles when minimap is off

src/misc.asm
```diff
@@ -32,6 +36,22 @@ org $828BB3
     JSL gamemode_end
 
 
+; Replace unnecessary logic checking controller input to toggle debug CPU brightness
+; with logic to collect the v-counter data
+org $828AB1
+    %a8() : LDA $4201 : ORA #$80 : STA $4201 : %ai16()
+    LDA $2137 : LDA $213D : STA !ram_vcounter_data
```
idlechild, line 43: This logic was moved.  In the new location it is overwriting a rep 30 (that's why it now has %ai16 instead of %a16), so that saves one instruction.  I also removed an AND #$00FF since that only needs to be done if we draw the CPU usage.

This saves 6 CPU cycles every frame

src/misc.asm
```diff
@@ -32,6 +36,22 @@ org $828BB3
...
+
+    ; For efficiency, re-implement the debug brightness logic here
+    LDA $0DF4 : BEQ .skip_debug_brightness
```
idlechild, line 46: This logic is reimplemented because it was moved up.  The previous overwritten logic would first check debug mode, and if in debug mode then check if the debug shortcut for brightness was input.  That's removed since we can just turn on debug CPU brightness through the menu.

Skipping the initial debug mode check saves 7 CPU cycles every frame

src/misc.asm
```diff
@@ -72,26 +92,33 @@ hook_unpause:
...
-    LDA !sram_artificial_lag : BEQ + : ASL #4 : TAX
-    {
-        - DEX : BNE -
-    }
-    +
+endif
+
+    ; If minimap is disabled or shown, we ignore artificial lag
+    LDA $05F7 : BNE .endlag
+    LDA !ram_minimap : BNE .endlag
```
idlechild, line 112: This LDA : BNE adds 6 CPU cycles every frame when minimap is off

idlechild: Thus overall, if looking at lag every frame when minimap is off, I think we only reduced lag by one CPU cycle.  Goal was to not change lag so I'm happy with that result.

src/infohud.asm
idlechild: Talked with Firebat and we think this is a non-issue.  I'm going to proceed with the merge.

It wouldn't surprise me to find an JML/JSL statement that is inaccurate for PAL; we still have more work to do for PAL.  For example, shinetune isn't up to date for PAL.

PR #47, opened 2021-07-07
Room/Door Timers with Minimap + Magnet Stair Fix

InsaneFirebat:
This splits `ih_update_hud_code` based on the minimap toggle. When enabled, the map tile counter will be drawn at the top-left where item% would normally be. The door transition timer is drawn directly below it. The room timer is drawn next to the minimap.

I also added a toggle for fixing the "magnet stairs" bug. If you're not familiar, the top of the stairs should have a square slope tile directly after it, but these are missing in the Magnet Stairs room. When your X position lands on the right value passing over them, you will come to a full stop for one frame.

This fix uses door asm (runs about halfway through transition) to overwrite the two tiles. I hijacked both doors just to prevent the inevitable "this doesn't work" when someone tries to test it before the escape. The menu toggle will also fix the stairs if you're currently in the room.

Edit:  Fixed a typo and tested the squash thing to merge the relevant commits. Worked well but doesn't seem to show what I did.

### DIFF
--- a/src/defines.asm
+++ b/src/defines.asm
@@ -36,6 +36,7 @@
 !ram_vcounter_data = $7FFB8C
 !ram_minimap = $7FFB98
 !ram_shot_timer = $7FFB9E
+!ram_magnetstairs = $7FFBA6
 
 !ram_kraid_rng = $7FFB78
 !ram_phantoon_rng_3 = $7FFB7A
--- a/src/infohud.asm
+++ b/src/infohud.asm
@@ -374,12 +374,84 @@ ih_update_hud_code:
     ; Bank 80
     PEA $8080 : PLB : PLB
 
-    LDA !ram_minimap : BEQ .start_update
+    LDA !ram_minimap : BNE .minimap_hud
+    BRL .start_update
 
-    ; Map visible, so draw map counter unless shinetune is enabled
-    LDA !sram_display_mode : CMP #$0007 : BEQ .minimap_end
-    LDA !ram_map_counter : LDX #$00B0 : JSR Draw4
-    LDA !IH_BLANK : STA $7EC6B8 : STA $7EC6BA
+  .minimap_hud
+    ; Map visible, so draw map counter over item%
+    LDA !ram_map_counter : LDX #$0014 : JSR Draw3
+    LDA !sram_display_mode : CMP #$0007 : BNE .minimap_roomtimer
+    BRL .map_doorlag
+
+  .minimap_roomtimer
+    LDA !sram_frame_counter_mode : BNE .ingameRoomMap
+
+    ; Real time with minimap
+    {
+        ; Divide real time by 60/50, save seconds, frame seperately
+        {
+            STZ $4205
+            LDA !ram_last_realtime_room : STA $4204
+            %a8()
+            if !FEATURE_PAL
+                LDA #$32
+            else
+                LDA #$3C
+            endif
+            STA $4206
+            PHA : PLA : PHA : PLA
+            %a16()
+            LDA $4214 : STA !ram_tmp_1
+            LDA $4216 : STA !ram_tmp_2
+        }
+        ; Draw seconds
+        LDA !ram_tmp_1 : LDX #$00B0 : JSR Draw2
+
+        ; Draw decimal seperator
+        LDA !IH_DECIMAL : STA $7EC6B4
+
+        ; Draw frames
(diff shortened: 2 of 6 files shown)


idlechild (commented):
Minimap changes look good.  I'll have to try them out, but I think this is useful.  I probably should have tried harder to make stuff fit; thanks for doing that!

Fixing magnet stairs... oof I am not a fan.  Generally the practice rom should be for practicing scenarios that can occur in game.  Fixing RNG to practice highly improbable events is fine, but I don't think we should extend to the impossible.

My understanding is magnet stairs occurs because the wrong tiles were used, so if samus is running on the stairs and hits the tile with certain pixel/subpixel position, she loses momentum.  If subpixel position were RNG dependent, I'd have no problem with fixing the RNG, but the position isn't RNG dependent.

I'd be a little more on board if the tile was obviously modified.  For example, the practice rom allows you to make Samus invincible, which is of course impossible in game without debug mode, but it's also fairly obvious when this mode is enabled.  These modified floor tiles should also be obvious.

Even then, not sure how popular magnet stair fix will be.

InsaneFirebat:
Interesting take on the stair fix. This is something I've actually been doing on a separate rom to practice stair room for years. It's really nice to be able to practice the optimal situation every time. It's nowhere near improbable to get "good stairs" in a run. I'd estimate that it happens about 30-40% of the time. For players that don't jump the stairs (and want top Ceres times), it's very worthwhile practice in my opinion.

I normally modify the appearance of the tile in my rom, but that's because it's otherwise vanilla and I don't want anyone playing on it without it being obvious. For the practice hack, I didn't think that was necessary because of the HUD. I included it as an option so that I wouldn't be distributing my "Ceres Training ROM" anymore to the hardcore Ceres grinders. I don't expect it to be popular, but I do expect it to be used by some top players.

InsaneFirebat:
I looked into updating the tile graphics and the room scrolling prevents updating both tiles at once, since only one is loaded into VRAM unless the screen is scrolled far enough. On top of that, VRAM can only be accessed during a blanking period, so I don't think I can do that from the menu toggle either.

I don't think it accomplishes anything, personally. The HUD already gives away that the ROM is modified, so you can't use it to cheat a run anyway. At best, you could use it to get the best room time, but the same could be done with a savestate in the door and memorizing the number of pumps needed to pass by.

idlechild:
Thanks for the insights.  I'm more on board today than yesterday.

I was thinking the tile modification may give some advantage for room times, but I forgot you could just armpump faster for more advantage.  Even if you had an "impossible" path through the room, where in the real game you would always get magnet stairs, if you threw in one more armpump then you'd likely avoid the magnet stairs and get the same or faster room time.

Another claim I made about RNG... actually RNG may play a significant role.  If Samus is between tiles and the game has to pick one, half the time it's going to pick the good slope tile you are on and not the bad tile you are approaching.

I think modifying the tiles would still be a good idea; less controversial that way.  My understanding is if you modify VRAM outside of a blanking period, then you get a black streak in your gameplay for one frame, which I don't think is a big deal, and decent chance the tiles will get modified within the blanking period unless you do it from the menu.  Definitely no one cares about a black streak on the menu for one frame.  I don't think the menu checks the blanking period when it overwrites or restores the infohud tiles.

idlechild:
All that said, if modifying the VRAM at runtime means we have to add conditional logic that has to be run every room, that's not good.  There's plenty of CPU bandwidth in the magnet stairs room, and probably also at the door transitions to magnet stairs room, but we shouldn't make a change that could impact CPU in other rooms.

InsaneFirebat:
I'll look into the VRAM stuff later today. If we have to execute code during gameplay, we could just add it to the room asm (jump pointer loaded from level data) to keep it out of the rest of the game.

InsaneFirebat:
It would also be trivial to mark the HUD in some way. We could rewrite a static value like Samus' HP and replace the high bytes with #$3C to change the colors to red/orange. The missile/bomb icons are expected to be empty in all but NewGame+, so we could write whatever's clever in that space. Door asm can restore the HUD before the next room.

InsaneFirebat:
I've decided that this needs to go in without a graphics change, or be pulled out entirely. It's not worth the trouble messing with VRAM and I don't want the tile marked anyway. I'm just going to remove that code from my custom builds and then there's really no point in doing it to the main branch. Let me know which you prefer.

PR #48, opened 2021-07-07
NTSC to PAL pseudo-patch

idlechild:
I thought it'd be a good idea to document the differences between NTSC and PAL, so if we're injecting code somewhere for NTSC we can look up the PAL equivalent.  Here is the progress I made so far... bank 82 really dragged on... I suspect I will continue but not sure how long it will take.


idlechild:
I took a different approach and made more progress, but still a ways to go before it is complete.  I think all of the inserts and removes are done so at least the general mapping is done.

I also found PJBoy's note on the incorrect Zebes planet tile.  I still need to test the fix but figured why not include it.

PR #50, opened 2021-07-15
Reorganize Defines

InsaneFirebat:
- Reorder defines.asm to organize our RAM map
- Replace !ram_ctrl1 with !IH_CONTROLLER_PRI
- Share !ram_HUD_check for simple HUD modes
- Replace !ram_max_etanks with !ram_etanks
- Fix Etank counter when vspeed displayed

I've grouped up related RAM addresses and left some gaps between so we have an easier time inserting new ones and finding the next available address. Nearly all of the current addresses are in $B00-BFF, with the menu specific RAM and stack in F00+. I have a personal request that !ram_cgram_cache have a 30 byte allotment so my custom builds can keep all their master branch defines untouched. I'm using 20+ bytes here to customize more of the menu palettes.

Gamemodes.asm was using !ram_ctrl1 and !ram_ctrl1_filtered where everywhere else used !IH_CONTROLLER_PRI. They pointed to the same addresses ($8B and $8F) so I made it consistent and removed !ram_ctrl1. You may notice the savestate defines also have $8B and $8F defined with different names. I left those alone since the savestate code is meant to be ported to other games. (should they be moved into save.asm instead?)

There were a few very basic InfoHUD modes that store the drawn value to check if it needs to be redrawn next frame. These now use a common address (!ram_HUD_check) to store the drawn value. Couldn't think of a better name for it. Please suggest.

The ?? comments on !ram_etanks and !ram_max_etanks were likely because the max value was only ever zeroed and never used otherwise. It's gone now and only !ram_etanks is used.

I made a mistake when setting up the branch to skip item% code when Vspeed is enabled. It will now calculate etanks before trying to draw the value.

### DIFF
--- a/src/defines.asm
+++ b/src/defines.asm
@@ -63,26 +63,26 @@
 !ram_roomstrat_counter = $7FFB82
 !ram_roomstrat_state = $7FFB84
 !ram_enemy_hp = $7FFB86
-!ram_mb_hp = $7FFB86
-!ram_shot_timer = $7FFB88
-!ram_shine_counter = $7FFB8A
-!ram_dash_counter = $7FFB8C
-!ram_shine_dash_held_late = $7FFB8E
-!ram_shinetune_early_1 = $7FFB90
-!ram_shinetune_late_1 = $7FFB92
-!ram_shinetune_early_2 = $7FFB94
-!ram_shinetune_late_2 = $7FFB96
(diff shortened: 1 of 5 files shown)


idlechild:
Look great.  Organizing the RAM addresses we use was overdue, and helpful to see where some free space is.

idlechild (approved):

PR #51, opened 2021-07-20
Preset scrolling fix + more early HUD updates

InsaneFirebat:
Just a couple more places to update the HUD early: Ceres escape elevator and facing forward on the ship. The ship update runs anytime you press down to go in the ship, not just the escape. Ceres elevator didn't need a separate address for PAL.

### DIFF
--- a/src/infohud.asm
+++ b/src/infohud.asm
@@ -140,6 +140,16 @@ org $AAE582      ; update timers when statue grabs Samus
 endif
     JSL ih_chozo_segment
 
+org $89AD0A      ; update timers when Samus escapes Ceres
+    JSL ih_ceres_elevator_segment
+
+if !FEATURE_PAL
+org $A2AA38
+else
+org $A2AA20      ; update timers when Samus enters ship
+endif
+    JSL ih_ship_elevator_segment
+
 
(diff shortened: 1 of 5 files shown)


InsaneFirebat:
I forgot to change the overwritten JSL address for PAL. Fixed now.

InsaneFirebat:
I've got a fix for the janky scrolling when loading Phantoon presets (among others) that I'd like to include in this pull. Just needs a bit more testing.

InsaneFirebat:
The scrolling fix first checks the 5th scroll slot to see if it's red (zero) and changes it to blue if so. This fixes a few different presets including Phantoon, GGG, and Aqueduct. It then checks for certain rooms that need a separate or addition fix. These fixes can be found in the door asm leading into the spot with a bad scroll.

InsaneFirebat:
This next commit resets the room timers when starting a new game in Ceres. Time spent in the title menus was included in the room times prior to this change. The hijack is right before the gamestate is incremented from 7 to 8.

Acid Chozo room (before GT) added to preset_scroll_fixes.

UNHchabo (approved):

PR #52, opened 2021-07-29
Auto Enemy/MB HP + Metronome + Preset Fixes

InsaneFirebat:
Door asm has been added to check for Enemy HP and swap it to MB HP if enabled. A few lines in ih_after_door_transition will check for MB HP enabled and swap it to Enemy HP if not in Mother Brain's room.

The metronome feature is far from perfect but could still be useful when learning the timing of something new. Sometimes the sound will "skip a beat" if other sounds are playing, so it's not going to be accurate if used during gameplay. The "BEEP" sound effect had it's priority increased to help with this, but other high priority sounds may still interfere. The beep (from scrolling the map when paused) was the only suitable sound I could find that wouldn't occur during normal gameplay, so it was the only one with a priority change. Making copies or adding new sounds isn't as straight-forward as I'd hoped since it's all wrapped up in the sound engine code.

There were a couple bugs with loading presets while Samus is "locked" in place. These are fixed by setting Samus' movement pointers to "normal" and setting the elevator status to "inactive".

The preset_scroll_fixes routine no longer applies to all rooms with red scrolls in the 5th slot. Only room specific fixes are applied now.


InsaneFirebat:
I added `!ram_game_loop_extras` to eliminate the timing change from adding the Metronome. GameLoopExtras is called when enabling features that rely on `ih_game_loop_code`, and again when exiting the menus. It checks if any of those features are enabled and toggles the flag. If the flag is zero, `ih_game_loop_code` will skip the extra features and jump to handling inputs.

InsaneFirebat:
About a year ago, I put in a fix for murder beam being equipped when selecting GT Code from the equipment loadouts. I didn't have a good understanding of what I was doing at the time, but it did prevent crashes. Looking back now, I can see why it only worked on the 100% loadout, and I wanna fix it right. Now it won't blindly unequip Spazer when it's not necessary.

I added a check for zero on the Load Last Preset shortcut. If your SRAM was fresh, it would load zeroes from the last preset pointer and crash. Now it jumps to random_preset instead of crashing.

InsaneFirebat:
I finally tracked down the new credits crash. `!ram_slowdown_mode` needs to be in $7E RAM space because the credits sequence will DMA the same junk data all the way to the end of $7F. 

Here's where I broke it on lines 14 (red) and 20 (green):
https://github.com/tewtal/sm_practice_hack/pull/50/commits/2ab74fd7424912246f44b3a943beb21bd8bdf294#diff-2f7631c3809de0a8eed06ae13581e360e7e8b4ef50328a16e3380b0503eae514L14

idlechild:
This all looks useful.  Good idea to optimize the extra game loop features to keep all of this CPU neutral unless you turn on the extra features.  Also good find on the credits issue.

PR #57, opened 2021-09-14
[2.2.8] New Presets and Fixes

InsaneFirebat:
Added GT Max% and Nintendo Power category presets. Max% presets follow the WR route and were organized by Ruyarin. Nintnedo Power presets were almost entirely made by princesskira69.

Added a menu option and controller shortcut to automatically reset the segment timer upon exiting the next door transition. This gives a more consistent timing for repeating segments.

Added another preset scrolling fix for Oasis (under Toilet) and made slight adjustments to PRKD presets.

Added menu options to adjust current energy and reserves.

Fixed a bug that prevented teleporting after a map station was used. This was caused by the `maybe_trigger_pause_long` routine that runs when exiting the menu. It jumps to a vanilla routine that sets the gamemode to Ch (pausing) if the message box index is 14h (map station). I'm not sure why this vanilla routine is being executed. It doesn't do anything unless the message box index is set to map station or save confirmation, neither of which are relevant to the practice menu. http://patrickjohnston.org/bank/85#f80FA

@idlechild Please add in your Grapple/Xray icon change :)


src/mainmenu.asm
```diff
@@ -355,6 +369,9 @@ eq_goto_toggleitems:
 eq_goto_togglebeams:
     %cm_submenu("Toggle Beams", #ToggleBeamsMenu)
 
+eq_currentenergy:
+    %cm_numfield_word("Current Energy", $7E09C2, 0, 1499, 1, #0)
```
idlechild, line 373: Max HP should be 2100

src/mainmenu.asm
```diff
@@ -370,6 +387,9 @@ eq_setetanks:
         STA $09C4 : STA $09C2
         RTS
 
+eq_currentreserves:
+    %cm_numfield_word("Current Reserves", $7E09D6, 0, 400, 1, #0)
```
idlechild, line 391: Max Reserves should be 700

src/defines.asm
```diff
@@ -218,10 +219,11 @@
 !sram_ctrl_kill_enemies = $702004
 !sram_ctrl_full_equipment = $702006
 !sram_ctrl_reset_segment_timer = $702008
-!sram_ctrl_load_state = $70200A
-!sram_ctrl_save_state = $70200C
-!sram_ctrl_load_last_preset = $70200E
-!sram_ctrl_random_preset = $702010
+!sram_ctrl_reset_segment_later = $70200A
```
idlechild, line 222: There should be init.asm changes to increment SRAM_VERSION and also initialize sram_ctrl_reset_segment_later to 0

idlechild:
Looks good (assuming we fix the areas I made comments on).

I'm surprised you didn't move the banks, weren't you planning on moving banks for better alignment with hacks?  We should at least move out of bank FF so SpriteSomething has a chance to work.

InsaneFirebat:
I forgot to move those orgs again after I remade my pull branch. I tested the hitbox patch, but it only works if SpriteSomething is patched first. This probably means the website won't allow it due to the new CRC checks. The menu and SRAM fixes are in as well. Thanks.

idlechild:
I can tack on my xray/grapple icon change before merging.

I was also considering throwing this graphic patch in, it's completely unnecessary but also why not?

[Imgur](https://i.imgur.com/GBUwHyj.png)

idlechild:
Unfortunate that SpriteSomething has to be patched first, wonder if that's something they could fix on their end.  It doesn't feel as compatible as they claim:
https://github.com/Artheau/SpriteSomething/blob/master/resources/app/snes/metroid3/FAQ.md

InsaneFirebat:
It seems to be overwriting everything with 00 bytes instead of just writing what it needs in the later banks. It crashes almost immediately when NMI tries to jump to InfoHUD code and executes BRK $00

idlechild:
So when their readme says "SpriteSomething will place tile data into banks $F5 through $FF", that's just bullshit since we aren't using those banks at all anymore, but you're saying they wrote 00s in other banks.

Okay so you have to run SpriteSomething on the vanilla rom first, and after you do that you can't get the automatic download from https://smpractice.speedga.me/?  Is that the issue?

Also okay with the zebes graphics patch?

idlechild:
They probably implemented the "expand from 3MB to 4MB" logic in a dumb way

InsaneFirebat:
Yes to all of it. Anyone looking to patch hacked roms will need to download the source or acquire an InfoHUD patch somehow. I was thinking about an option on the website to "patch at your own risk" and skip the CRC check, but I wouldn't know where to start with implementing that. I might look at it a bit closer soon.

idlechild:
I was also thinking "patch at your own risk" except just making the .ips file downloadable (or maybe that's what you meant)

Then I started thinking SpriteSomething should just fix their issue, and if that happened then we don't have to make website changes; it's nice as is.  I'm going to ask for assistance on the SRL - Super Metroid discord (as long as it is still called that; there's talk of renaming it)

You know you're always welcome to come back :)  There's slightly less content since ED got booted a week or two ago.

idlechild:
Huh I was hoping this commit would auto-close.  Maybe I should have done a merge instead of rebase (although rebase feels more correct).  Well whatever, it's merged.  Closing this PR.

InsaneFirebat:
I just tested with the latest version of SpriteSomething and it appears this isn't an issue anymore. I can patch in either order and it works. Sorry for the false alarm.

PR #59, opened 2021-09-20
[2.2.9] Custom Presets with SRAM

InsaneFirebat:
Custom presets allow users to save data to SRAM and load it in place of the addresses and values from (category)_data.asm files. When loaded, a flag is set that is checked just before `!ram_load_preset` and jumps to `preset_load` if non-zero. When it's time to load the category data-chain, it checks the flag again and skips to `custompresets.asm` code. Bank $70 is the only part of SRAM that we can use freely, and we typically use SRAM in the $702xxxx range, so the preset slots start at $703000 and run almost to $708000. That gives 47 slots for users to play with and leaves way more SRAM free than we'll need for normal features.

All of that comes with four new controller shortcuts:  Save Custom Preset, Load Custom Preset, Next Preset Slot, and Previous Preset Slot. Changing the slot with controller shortcuts will display the slot number next to Xray on the HUD, until it's overwritten with a blank tile later.

The controller shortcuts are starting to pile up, especially in my custom builds, so I separated them into two groups. The first group of important shortcuts are always checked (when there are new inputs) and the other group will only be checked if at least one of them is configured. `GameModeExtras` is run when configuring a shortcut in the menu and during boot just after SRAM init has passed. This sets the flag if any less common shortcuts are configured.

PRKD presets now come with an option for No-X Ridley strats, which keeps Ice and Spazer equipped in Lower Norfair until Ridley is defeated. Some runners have also requested the Business Center preset start before the elevator so I've added that in as a new preset in the Red Brinstar Revisit section. Scrolling fixes for bottoms of Crab Shaft and WS Shaft are included. More of the item selections for various presets as well.

I moved the `incsrc preset_menu.asm` stuff from mainmenu.asm to presets.asm. I also trimmed out a lot of the debug text so it just says where the data ends for each bank.


InsaneFirebat:
Added in category presets for 100% Map Completion by mm2nescartridge. With new presets comes new scroll bugs, so those are fixed too. The room checks are ordered by room ID so I can compare it at the start and cut the list of potential checks in half.

src/custompresets.asm
```diff
@@ -0,0 +1,460 @@
...
+    dw $0000, $01AE, $035C, $050A, $06B8, $0866, $0A14, $0BC2
+    dw $0D70, $0F1E, $10CC, $127A, $1428, $15D6, $1784, $1932
+    dw $1AE0, $1C8E, $1E3C, $1FEA, $2198, $2346, $24F4, $26A2
+    dw $2850, $29FE, $2BAC, $2D5A, $2F08, $30B6, $3264, $3412
+    dw $35C0, $376E, $391C, $3ACA, $3C78, $3E26, $3FD4, $4182
+    dw $4330, $44DE, $468C, $483A, $49E8, $4B96, $4D44
+
+custom_preset_save:
+{
+    LDA !sram_custom_preset_slot : ASL : TAX
```
idlechild, line 17: If we allocated 512 bytes or 200h of space for each preset, that would allow us to add a bit more stuff in the future without forcing a compatibility break.  It might also be easier to reuse this space in the future, in case someone wants 20 custom presets and 20... whatever the next cool feature is.

We could also get rid of the PresetSlot table since we could just ASL : XBA : TAX to multiply !sram_custom_preset_slot by 200h.

src/custompresets.asm
```diff
@@ -0,0 +1,460 @@
...
+    LDA.l PresetSlot,X : TAX : LDY #$0000
```
idlechild, line 18: Why LDY #$0000?

src/custompresets.asm
```diff
@@ -0,0 +1,460 @@
...
+    LDA #$5AFE : STA $703000,X : INX #2 ; mark this slot as "SAFE" to load
```
idlechild, line 19: After this slot, the next slot would indicate how much space is actually used by the preset.  You had that at 1AEh but since we'd add two more bytes it would return 1B0h.  Theoretically we could use that information to selectively load only some of the information.

Right now we'd probably just ignore it with an INX #2 when loading a custom preset.

src/custompresets.asm
```diff
@@ -0,0 +1,460 @@
...
+    LDA $7ED91C : STA $703000,X : INX #2 ;  Events, Items, Doors
+    LDA $7ED91E : STA $703000,X : INX #2 ;  Events, Items, Doors
+    RTL
+}
+
+custom_preset_load:
+{
+    LDA !sram_custom_preset_slot : ASL : TAX
+    LDA.l PresetSlot,X : TAX : LDY #$0000
+    INX #2 ; skip past "5AFE" word
```
idlechild, line 241: After this slot, the next slot would indicate how much space is actually used by the preset.  You had that at 1AEh but since we'd add two more bytes, we'd expect 1B0h.  Theoretically we could use that information to selectively load only some of the information, in case we add more information in the future to custom presets.  So in the future if we're storing 1C0h of info, then when loading an older custom preset with 1B0h info, we'd know to skip over the last 16 bytes of information.

Right now we'd probably just ignore it with an INX #2 when loading a custom preset.

src/custompresets.asm
```diff
@@ -0,0 +1,460 @@
...
+    LDA $0913 : STA $703000,X : INX #2 ;  Screen subpixel Y position
+    LDA $0915 : STA $703000,X : INX #2 ;  Screen Y position in pixels
+    LDA $093F : STA $703000,X : INX #2 ;  Ceres escape flag
+    LDA $09A2 : STA $703000,X : INX #2 ;  Equipped Items
+    LDA $09A4 : STA $703000,X : INX #2 ;  Collected Items
+    LDA $09A6 : STA $703000,X : INX #2 ;  Beams
+    LDA $09A8 : STA $703000,X : INX #2 ;  Beams
+    LDA $09C0 : STA $703000,X : INX #2 ;  Manual/Auto reserve tank
+    LDA $09C2 : STA $703000,X : INX #2 ;  Health
+    LDA $09C4 : STA $703000,X : INX #2 ;  Max helath
```
idlechild, line 41: Fix helath typo (here and also below)

src/defines.asm
```diff
@@ -1,38 +1,39 @@
...
+!ram_last_realtime_room = $7FFB0A
+!ram_last_room_lag = $7FFB0C
+!ram_last_door_lag_frames = $7FFB0E
+!ram_transition_counter = $7FFB10
+!ram_transition_flag = $7FFB12
+!ram_transition_flag_2 = $7FFB14
+
+!ram_seg_rt_frames = $7FFB16
+!ram_seg_rt_seconds = $7FFB18
+!ram_seg_rt_minutes = $7FFB1A
```
idlechild, line 16: Lol just love moving stuff around eh?

So I know these RAM values are undocumented, but I think someone is actually pulling these three values (frames/seconds/minutes) out of the practice rom.  IIRC it's trying to track your best segment time... or maybe it's the room times up above.

Anyway, if we don't have to change these values, I don't think we should.  It's good we did it for that big reorg a little while ago where defines were very disorganized, but not sure we should keep moving them around.

web/data/config.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Super Metroid Practice Hack",
-    "version": "2.2.8",
+    "version": "2.2.9",
```
idlechild, line 3: If we do all of these changes at once, we might as well go to 2.3.0, especially since we should decrement the artificial lag for total's changes.

src/mainmenu.asm
```diff
@@ -202,71 +204,150 @@ mm_goto_ctrlsmenu:
...
-  incsrc presets/kpdr21_menu.asm
-  print pc, " kpdr21 menu end"
+presets_goto_select_preset_category:
+    %cm_submenu("Select Preset Category", #SelectPresetCategoryMenu)
 
-  print pc, " hundo menu start"
-  incsrc presets/hundo_menu.asm
-  print pc, " hundo menu end"
+presets_custom_preset_slot:
+    %cm_numfield("Custom Preset Slot", !sram_custom_preset_slot, 0, 46, 1, #0) ; update max slots in gamemode.asm
```
idlechild, line 221: Max slots would be adjusted to 39 per my suggestion (40 custom presets total)

src/custompresets.asm
InsaneFirebat: Leftover index from when I was saving/loading enemy data in custom presets. I pulled that out because it used way too much space, and I still wasn't tracking enough enemy slots for it to be used the way I envisioned.

InsaneFirebat:
Added in a little fix for calculating item%. It was previously counting an unused item bit set by GT Code.

src/gamemode.asm
```diff
@@ -70,32 +78,27 @@ gamemode_shortcuts:
     AND !IH_CONTROLLER_PRI_NEW : BEQ +
     JMP .full_equipment
 
-  + LDA !IH_CONTROLLER_PRI : AND !sram_ctrl_random_preset : CMP !sram_ctrl_random_preset : BNE +
-    AND !IH_CONTROLLER_PRI_NEW : BEQ +
-    JMP .random_preset
-
+  .check_menu
   + LDA !IH_CONTROLLER_PRI : AND !sram_ctrl_menu : CMP !sram_ctrl_menu : BNE +
```
InsaneFirebat, line 98: I force pushed the plus label back on this line to fix a bug where the menu shortcut would stop working.

PR #60, opened 2021-09-24
Hitbox and OOB viewers

tewtal:
This PR includes two changes.
 - The first is a simple sprite overlay that shows Samus current hitbox on top of Samus that can be toggled on or off.
 - The second is a simplified version of the OOB viewing Lua scripts for emulators. It will draw a grid of tiles on top of the screen as sprites that represent the clipdata of tiles nearby Samus, coloring the tiles differently depending on the clipdata block type. This should be useful for anyone learning OOB categories.


tewtal:
Not sure what other plans are being worked on right now so didn't want to merge this straight away. It'd be great it someone could check it out and make sure I didn't break anything since it's been a while since I added things to the practice hack.

InsaneFirebat:
This is really cool stuff. I'm surprised at how well it runs in some rooms, others not so much. Very impressive even with the lag. I've had many requests to do this in the past and I honestly didn't think it was realistic.

One small bug I found is that it stops drawing if you load a preset or teleport. It has to be toggled off and back on again to show up. Wouldn't be an issue except that it still lags like it's running even though it isn't drawing anything on screen.

tewtal:
Aha, it's likely that the custom sprite tiles gets unloaded when you load a preset or teleport. That explains why it still lags, because it's running and just drawing invisible sprites. I'll fix it!

src/spritefeat.asm
```diff
@@ -0,0 +1,330 @@
...
+    PHA : PHX : PHY : PHP
+    
+    ; Draw OOB viewer if activated
+    JSR update_sprite_oob
+
+    ; Draw Samus hitbox if activated
+    JSR update_sprite_hitbox
+
+    PLP : PLY : PLX : PLA
+    JSL $A0884D
```
InsaneFirebat, line 38: if !FEATURE_PAL
    JSL $A0885D
else
   JSL $A0884D
endif

Didn't find any other PAL conflicts.

tewtal: Oh right, PAL is a thing we do now. I completely forgot about that :)
Thanks for noticing, I'll add that in.

src/defines.asm
```diff
@@ -35,6 +35,9 @@
 !ram_map_counter = $7FFB38
 !ram_vcounter_data = $7FFB3A
 
+!ram_sprite_hitbox_active = $7FFB46
+!ram_oob_watch_active = $7FFB48
+!ram_oob_watch_lag_backup = $7FFB4A
```
idlechild, line 40: I don't think you used ram_oob_watch_lag_backup

src/spritefeat.asm
```diff
@@ -0,0 +1,336 @@
...
+;
+org $9ADAE0
+db $FF, $FF, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80
+db $FF, $FF, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80, $80 
+
+
+org $F0E000
+print pc, " spritefeat start"
+; This runs every frame before any other sprites are drawn, this is needed so we can get priority over everything else
+update_sprite_features:
```
idlechild, line 28: By my count, by default (new features off) this method adds 90 clock cycles per gameplay frame.  Adding in 8 cycles for the new check in the artificial lag check and we've got 98 cycles... that's a lot!

We should decrease the default artificial lag by one to get back 80 cycles.  I'd also recommend optimizing this method a bit.  At the beginning of the two JSR methods is a check, when false we RTS.  We should be doing that here and only JSR if we need to (saving a JSR and RTS each for 24 cycle savings total).  That would make the total CPU hit almost even.

src/spritefeat.asm
```diff
@@ -0,0 +1,336 @@
...
+update_sprite_oob:
+    LDA !ram_oob_watch_active : BNE +
+    RTS
++
+    !oob_width = $000D
+    !oob_height = $0009
+    LDA $590 : STA $C8
+
+    LDA $0AF6
+    SEC : SBC #(!oob_width*8)
```
idlechild, line 78: Thank you for this example of local constants.  This will make it easier for me to update the shinetune method so it works for PAL.

PR #61, opened 2021-09-27
Add sound effects to magic pants

NobodyNada:
Uses the configurable metronome sound effect. I'm a musician, so
practicing with an audio cue greatly improved my short charge.

If people find this annoying/unhelpful and want a menu option to turn it off, let me know and I'll add that.


NobodyNada:
(force-pushed to fix a typo without cluttering up the git history)

NobodyNada:
Gotcha! I'll try and get that implemented later today.

NobodyNada:
I've added a "Loud Pants" toggle that turns the sound effect on/off. The same toggle also adds sound effects to Space Pants, because that seems like the most consistent/least surprising thing to do. I also added a note in the documentation.

(If you have recommendations for a more intuitive UI, I'm all ears. I would rather have made the "Magic Pants" option toggleable between Off, Quiet, and Loud, but I did not find a good way to do this because the "Off" graphic requires a customized palette that didn't play nice with the  "choice" menu item type.)

InsaneFirebat:
We could optimize this a little more by eliminating the jump/return and just copying the SFX code into each pants routine. ROM space is basically a non-issue since the extra banks were added.

NobodyNada:
Sounds good, done.

And thanks! I've recently started playing/casually-pretending-to-speedrun SM and I have been thoroughly impressed by the quality (and sheer volume) of documentation and tooling you guys have built. I hacked this together because it helped me practice, so I figured maybe someone else would find it useful.

idlechild:
Looks good.  I tried this out today, it was helpful in updating shinetune to work for PAL.

PR #65, opened 2021-10-21
Combined Pull Request

idlechild:
The four other pull requests combined into one.  Also finished updating infohud modes for PAL (except moat CWJ, planning to do that one later).  Only thing left to do is update the help pages and then I think we're ready to go.

This does still include my kill enemies easter eggs.  Let me know if there are any objections.


idlechild:
I added a "Lag Frames" section to the help page

PR #66, opened 2021-10-31
Added room strat for the parlor-climb door skip

idlechild:
First draft.  I want to see if I can improve tracking during the door transition (also would be nice if I can make shortcuts work during the door transition).  However feel free to take this and build an early version of it (hint hint @InsaneFirebat, btw thanks again for the OOB tile viewer controller shortcut).

Interestingly if you are one frame late to expand hitbox, that actually gets detected and you can see Samus expand hitbox for that frame before hitting the door transition.  I'm not sure why that works but it makes this first draft a lot more useful.  If you hit the door transition and don't get feedback then you expanded hitbox at least two frames late.  This all assumes you time the jump correctly; if the first two characters aren't YY then you won't get any feedback on expanding hitbox.

To make this work, I did stop the practice rom infohud from updating while on the pause menu.  I've never liked that the status options operate during that time but I've also never had a reason to go after that until now.  The only practice rom infohud feature that will update while paused is Samus HP.


idlechild:
This is a better version of the door skip along with some other features.

Also added misc options for the elevator flags.  The bad yellow blocks appear to be elevator platforms, triggering the game to set a flag that samus is standing on an elevator, which causes samus to be frozen when you enter the next room.

I'd like to replace the max etanks display with icons that would tell you this, icon for elevator, icon for flash suit, etc.  Not sure how much work that will be though.

idlechild:
Alright I think 2.3.1 is ready, just going to let this sit for a couple days in case anyone has comments.  The bank $84 changes, RAM migration and spacetime patch to skip over our RAM are all interesting changes and worth reviewing if you haven't already.

I'm wondering if x-ray can still corrupt our RAM now that it is at 7E.  I'm not sure how the x-ray corruption works.

@UNHchabo curious if those white and blue boxes are gone now that we moved our ram from bank 7F to bank 7E.

idlechild:
Tacked on one more feature (rando suit properties option).  Now I just want to walk through a glitched category to see how it compares to vanilla and then this should be ready for release.

idlechild:
Estimated impact on CPU at runtime (every gametime frame):
-5 cycles (remove CMP : BEQ before incrementing gametime)
+7 cycles (add LDA : BEQ checking if status icons enabled)
-12 cycles (remove unnecessary JSR to and RTS from an empty routine)
+12 cycles (add LDA : ASL : PHA to set an index used later to jump to a damage check routine)
+22 cycles (replace JSR with PLA : PHX : TAX : JSR,X : PLX : NOP : NOP for jump to damage check routine)
-19 cycles (replace PLP : RTS : JSR : PHP : REP #30 with BRA to combine damage check and pause check routines)
-10 cycles (replace RTS : JSR with NOP to combine pause check and low health check routines)
+6 cycles (add three NOPs)

Result: Added one cycle (I was thinking I removed one cycle but now thinking I miscounted earlier)

Also the periodic damage routines may have different CPU.  It's probably faster on frames where the subdamage has no carry, but slower on frames where there is a carry.

PR #67, opened 2021-11-26
Practice Hack 2.3.2

idlechild:
Still a work in progress.  I want to incorporate fast menu scrolling (although not sure it will be as elaborate as what IFB has).  I also want to add a room strat for snail clip to indicate if the snail height is good or not.


idlechild:
Alright I think it is ready.  Planning to release it sometime Saturday unless there are objections.

InsaneFirebat:
I haven't been able to figure out why just yet, but the loadstate during fanfare commit allows the message box to be closed early when fanfares are enabled.

idlechild:
Great catch @InsaneFirebat thank you

The issue is I used the wrong rep/sep macros, resulting in the X/Y registers getting set to 8-bit mode instead of being left in 16-bit mode.  This caused the #$0168 frame count to be reduced to #$68, so required fanfare delay was reduced by 256 frames.  This only happens when fanfare music was on.  With fanfare music off the frame count is #$0020 so it's already an 8-bit value.

Tbh I was expecting the PHP/PLP to undo any rep/sep changes but apparently that is not the case.  I will need to be more careful with my rep/sep/macro calls in the future.

InsaneFirebat:
Ah, I guess that's why it wasn't immediately obvious from the trace.

InsaneFirebat:
Bad things happen to the menu when you try to change the load state shortcut. Menu is bricked if you use Clear Shortcuts at the bottom.

idlechild:
> Bad things happen to the menu when you try to change the load state shortcut. Menu is bricked if you use Clear Shortcuts at the bottom.

Well heck!

Yep I caused these issues.  They should be fixed now.  I forgot to cover the use case where there is no load shortcut set.  Also I didn't realize our menu was using the vanilla logic for waiting for lag frame, which means we were checking for the load shortcut while in the menu which we do not want to do.

PR #68, opened 2021-12-07
[2.3.3] More hitbox viewers

InsaneFirebat:
Some of this pull is straightforward, and some of it should probably be up for a little debate. The controversial stuff comes from the menu changes. Please feel free to shoot down individual changes as you see fit. I know they aren't going to please everyone :)

I've added footers to a couple submenus that just draw a line of unselectable text at the bottom. I'm currently using it to explain how to set controller shortcuts, to explain that Room Strat is connected to the HUD mode, and to warn of crashes with the glitched beams menu. I originally thought I would be using it in more places, but it turns out most of those places only applied to my custom builds. I still like having it though, even if it were only used on the controller shortcuts menu.

Support for blank lines in the menu was also added. This lets us add gaps in the menu that sorta help organize similar options into groups. The blank lines are automatically skipped by the cursor. Since I was already going over all the menus to add these blanks, I decided to do a little reorganizing. A few things were reordered, but the ones people are most likely to complain about are moving fanfare and music toggles to the Game menu.

I've adapted total's hitbox for Samus into a few other hitbox viewers. Now that we have a few of these sprite features ready (and at least one more planned from total) I wanted to give them a separate submenu, named Sprite Features. Since these new features add a few more checks every frame, I decided to gate them behind a common variable, like game_mode_extras and game_loop_extras. I also set it up to only push/pull A unless those features are enabled, and only preserve the other registers if needed. It ends up being a sizeable savings on CPU cycles. Should we consider burning them off with artificial lag?

- Enemy hitboxes:  Draws around the first 8 enemies/slots, Does not work on bosses or enemies that use the extended spritemap format. Slot count could be increased but usability becomes inconsistent depending on the room.
- Samus projectile hitboxes:  Draws around all 5 of Samus' projectiles/slots.
- Enemy projectile hitboxes:  Draws around all 18 enemy projectiles/slots.

Each of the new hitbox viewers has a small problem with wrapping at the edges of the screen. I think it's fixable, but not worth the extra work (on the CPU side).

Finally, we've got some small fixes here and there. The weird graphics bug when loading presets into Fireflea room (Springball Maze / Hota Special presets) is now fixed by aligning the camera position to the top edge of the room. I added a little routine to category selection to zero the last_preset when changing categories, which prevents a crash from a bad pointer. Kill Enemies was added as an option in Misc, but it can't open the message box routine or the menu is toast. The built-in debug option to disable enemy projectiles is also in the menu now, so I can finally practice Caterpillar room without getting shot by the cacs every time :D


InsaneFirebat:
I'll also change the address that's checked to skip artificial lag if the tile viewer is active, but I'm waiting to see if any other changes are needed before commiting.
https://github.com/tewtal/sm_practice_hack/blob/master/src/misc.asm#L291-L292

idlechild:
Good call on skipping artificial lag with your new variable instead the old one, I missed that in my review.

It looks good... the only thing is the CPU optimization is too good.  I estimate you saved ~30 CPU cycles by skipping the PHX/PHY/PHP/PLP/PLY/PLX and also an extra LDA/BEQ.  I suggest adding NOP #15 to balance it out.  It'll be nice in the future when we add stuff we can reduce the NOPs to compensate.

InsaneFirebat:
> I suggest adding NOP #15 to balance it out.

Can we add it to the artificial lag default instead? Seems like the more appropriate place for it, and we'll always remember where to find those cycles later. +5 should equal 30 cycles.

src/init.asm
```diff
@@ -126,7 +126,7 @@ init_sram_upto9:
     ; $0010 = R
 
     ; Features
-    LDA #$0015 : STA !sram_artificial_lag
+    LDA #$001B : STA !sram_artificial_lag
```
idlechild, line 129: Not what I had in mind; you don't want this change.  sram_artificial_lag gets multiplied by 16 (ASL 4) before being used as the loop variable.

I'm getting ready to pull your stuff into my branch so I can fix it then.  I just wanted to give you a heads up so you don't start defaulting the artificial lag that high before handing off to others in case the extra lag is noticeable.

InsaneFirebat:
Thanks, I probably should have read the comments again instead of just looking at the loop :)

That means a minimum of 80 cycles to alter artificial lag. The NOP string is starting to look less cringe now... (still letting you make the change)

InsaneFirebat:
Merged into #69

PR #71, opened 2021-12-10
Preserve layer 2 offsets in presets

idlechild:
This should allow existing presets and custom presets to work, but also allow new custom presets to maintain the BG2 layer positioning.  It's more noticeable in places like Wrecked Ship or Tourian (e.g. zeb skip preset).  Eventually we should also add BG2 to the existing presets but that may take some time; I'm planning to ask for help from the community.

This is hot-off-the-press.  I'd like to clean up the code a bit, add some comments, not sure how I feel about the custom preset logic loops I added but they work and it's less code.  I didn't like the INX 2 everywhere so I am (mostly) maintaining X where it is.  This also helps so I can look up the size (at beginning of data) when deciding if I want to load BG2 (from the end of the data).

Also unfortunately this is did not fix the bowling room.  In particular the Pancakes and Wavers preset, which looks wrong when you load it (background is dark).  You can exit/reenter the room and it fixes the background, but when you get to bowling it is messed up.  However if you load the Bowling preset, the bowling room works.  It's rather strange.  Even with my improvements, custom presets do the same thing when set at the Pancakes and Wavers position.

### DIFF
--- a/src/custompresets.asm
+++ b/src/custompresets.asm
@@ -7,445 +7,139 @@ custom_preset_save:
 {
     LDA !sram_custom_preset_slot
     ASL : XBA : TAX ; multiply by 200h (slot offset)
-    LDA #$5AFE : STA $703000,X : INX #2 ; mark this slot as "SAFE" to load
-    LDA #$01B0 : STA $703000,X : INX #2 ; record slot size for future compatibility
-    LDA $078B : STA $703000,X : INX #2 ;  Elevator Index
-    LDA $078D : STA $703000,X : INX #2 ;  DDB
-    LDA $078F : STA $703000,X : INX #2 ;  DoorOut Index
-    LDA $079B : STA $703000,X : INX #2 ;  MDB
-    LDA $079F : STA $703000,X : INX #2 ;  Region
-    LDA $07C3 : STA $703000,X : INX #2 ;  GFX Pointers
-    LDA $07C5 : STA $703000,X : INX #2 ;  GFX Pointers
-    LDA $07C7 : STA $703000,X : INX #2 ;  GFX Pointers
-    LDA $07F3 : STA $703000,X : INX #2 ;  Music Bank
-    LDA $07F5 : STA $703000,X : INX #2 ;  Music Track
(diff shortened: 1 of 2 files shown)


idlechild:
I turned off the scrolling sky asm when loading presets.  Upside:  Pancakes and Wavers preset no longer corrupts bowling.  Downside:  Presets at landing site and ocean look bad until you leave the room.

As bad as the landing site/ocean presets look... I'm inclined to make this change.  I really don't like how a preset in ocean room (before pancakes and wavers) can mess something up several rooms away (bowling) and if it is corrupting graphics then it's hard to say what else it might be messing up.

Perhaps we could only apply this change to ocean and leave landing site alone... but without knowing why ocean corrupts graphics, are we sure landing site is safe?  Custom presets are still relatively new; perhaps there are some preset locations in landing site that would cause problems and we just haven't found them yet.

It would be nice if we could fix the dumb green box that appears.  Try some presets with this change and you're sure to see it.  If anyone knows how to get rid of that I'm interested.

idlechild:
Okay this version looks better... except the green box is larger.  We have to get rid of that.  Help!

Might be worth posting a short video of it on metconst and see if they know how to remove it.

idlechild:
This latest change is just for demonstration.  If the sky scrolling is disabled, only half of the landing site and west ocean background is updated and the other half is uninitialized.  So if we can just fix that, the green boxes issue will go away.

idlechild:
Yes!  One more change it's working much better now.  Sure the sky doesn't scroll when load a preset in one of those rooms, but it's a minor thing, and normally you're just going to leave the room anyway.  Sky still scrolls normally in the demos.  I need to test entering Zebes from Ceres or loading save at the ship, but assuming those work this is good to go.

InsaneFirebat:
Looking good now. Scrolling sky works perfect if you enter the room the proper way (from Ceres, door, or ship load) and a lot of the other random background bugs seem fixed or much less intrusive. Nice work!

idlechild:
Yep yep, planning to roll all the PRs into one today, might even release later today.

PR #73, opened 2021-12-15
Add layout menu

idlechild:
Starting with Firebat's magnet stair option, and also putting the menu footer to good use.  Next I plan to add area rando layout, and the varia anti-softlock patch after that.  I'm not planning to merge this in until that's all done.


idlechild:
It's a start.  More rooms to come for the area rando patch.

idlechild:
Added an option to enable PJBoy's fix for scrolling alignment/misalignment.

I was expecting this to affect how presets load, but apparently we don't use the method he modified.  I haven't spent much time with this option enabled yet, but thought I'd make it available.  Of course it should be defaulted off so that, if you do something that would cause the misalignment in the actual game, then you see the misalignment in the practice rom.

idlechild:
Thanks InsaneFirebat, stole at least half the low health alarm logic.  I had no idea I needed to modify three places instead of just one (not sure I've tested those other places yet either...)  I added a couple more options in case you want to keep the alarm but you want it to work properly.  Of course the default is still the vanilla behavior.

idlechild:
The new maridia save stations don't actually work; loading takes you to other areas in maridia.  I copied the patches from VARIA, but there's probably another patch to enable saving/loading to debug slots.  It's not worth fixing; in fact it is nice since there won't be any strange condition where you save with area rando on and load with area rando off.

InsaneFirebat:
You would need a separate patch to make save stations use more than the first 8 slots. I think you just need to change the `AND #$0007` at http://patrickjohnston.org/bank/84#f8CF1 ($8D0C) if you want it to work. I tested it Crab Shaft by manually setting the save data to that load point and it works fine with and without the area rando setting enabled.

idlechild:
Thanks IFB for the tip and trying it out.  It's a little funny but it works.

The maridia warp door was less work than expected.  Deer Force left a door unused to make things even easier.  The graphics are a little off when you use the warp door to enter the room that doesn't normally have a door.  I'm guessing the door transition starts before the asm inserts the door tiles or something like that; probably a lot more trouble to fix than it is worth.

Area Rando patch is basically done if anyone wants it.  The only thing left is Wrecked Ship doors when Phantoon is still alive, but that's something also affected by the VARIA tweaks and possibly the anti-softlock patches, so I will start working on those.

idlechild:
^^ THIS ^^
Save a preset in Ceres Ridley room with timer running, and loading the preset triggers the escape timer
Save a preset in MB room with glass broken, and loading the preset sets MB1 health to 0 in order to trigger MB2 fight
MB3 and Zebes Escape preset support is coming, along with the faster MB cutscenes from VARIA

Spent longer than I wanted figuring out ceres escape, and it's still not quite right.  I don't know where the background red shading is coming from, and I also don't know why the door isn't being drawn during the timer setup.  For now I'm going to move on to the MB presets and cutscenes.

InsaneFirebat:
Very nice. I've been testing out ways to start the Zebes escape on and off but haven't gotten far enough to put anything together yet. It'll be nice to have the Ceres escape starting properly too.

You can fix the Ceres Ridley's door by unsetting the #$0100 bit at $0FC6. Unfortunately, it needs to happen sometime after the preset routine is finished.

idlechild:
Thanks for the door tip.  I should be able to make it work now that I know what flag to set.  It'd be nice to figure out the red background thing, but that isn't as bad as the door.

I have Zebes escape working locally... sort of.  I have to cheat by setting the zebes ablaze flag myself, but it shouldn't be hard for me to change the code to do that.  With that flag set (and glass broken flag set) and load preset in MB room, MB1 dies and it goes to the escape... but MB1 is still sitting there and glitchy (and you take damage if you bump into it).  I'll add it to the PR when I have a better version.

idlechild:
Now we're cookin'

@UNHchabo, I was curious how the preset updates are coming.  I plan to replace the ceres last 3 doors preset with a ceres escape preset.  It might be easier if I did that on top of the BG2 updates instead of having to merge it later.  I don't need it tomorrow; I'm still not that close to finalizing this pull request.  Just a reminder.

idlechild:
Of course as soon as I post this, I managed to break the sound engine even with these changes.  I was loading state while in the middle of a door transition where music changes.  I thought I had that covered but I'll have to think about this some more.

There's at least one savestate I had that went from crashing almost every time to not crashing with these changes, so some progress.  Hopefully the more common use case of saving at a bad time (while music is changing) and then loading at a normal time is fixed.

idlechild:
Whew looks like it was just a dumb mistake, I didn't mark the music data as music data.  Gaining confidence in this solution to the music crashes.

PR #74, opened 2021-12-19
[2.3.4] Extended Spritemap Hitbox Viewer

InsaneFirebat:
The extended spritemap hitbox viewer is mostly useful for Phantoon and Draygon, but also works for space pirates and a few other enemies that didn't show correctly with the normal enemy hitbox viewer. Mother Brain, Ridley, Crocomire, and Kraid are still not supported. I'm told they (at least MB and Kraid) use yet another custom hitbox format. Yay...

RAM Watch can now handle 8-bit writes. Not much to say there. It uses the low byte as you would expect.

All of the presets I've made except All Bosses (who uses those anyway) have been updated with BG2 data. I also threw Ceres back in a few categories, cause it was in the original presets/movie. All or most of those movies are stashed in the tools folder on my fork in case someone ever needs them.

I also did the thing with Botwoon's RNG control. The bits are set in the menu now instead of during gameplay.


src/mainmenu.asm
```diff
@@ -1696,31 +1709,37 @@ ramwatch_lock_right:
 action_ramwatch_edit_left:
 {
     LDA !ram_watch_left : TAX
-    LDA !ram_watch_bank : BEQ .bank7E
+    LDA !ram_watch_write_mode : BEQ +
+    %a8()
++   LDA !ram_watch_bank : BEQ .bank7E
     CMP #$0001 : BEQ .bank7F : BRA .bankSRAM
```
idlechild, line 1715: I doubt this works in 8-bit mode.  I expect the SNES to interpret CMP #$0001 as CMP #$01 followed by a 00 BRK instruction.

idlechild: Also I've never done this before
Also don't look at this commit nothing to see here, this obviously fixes nothing
https://github.com/tewtal/sm_practice_hack/pull/73/commits/339dfcbe21c2efd7acdbfbbc8cf9c1216b852067

idlechild: Also I could be wrong, programming is fun

InsaneFirebat: Interesting. I wonder if there's something that makes it compile differently sometimes. It's working fine on my custom build but crashes in this version. I'll change the `CMP #$0001` to a `DEC` to fix it.

InsaneFirebat:
New commit adds some graphics I've been meaning to pull for a while. The + is needed for my new hitbox option (Phan + Dray Hitboxes) and the (parenthesis) were being used in RBO presets and displaying incorrectly. New stuff is circled here: https://i.imgur.com/DJ0QSEE.png

The 8-bit crash fix is in there too.

InsaneFirebat:
Added the reserve energy counter with option to display that or item%, still defaulted to item%.

idlechild:
I think I'm going to take IFB's PR here, add Ceres Escape preset support and the music fixes, push it and call it an interim release.  I want more time for the room layout changes, and fast MB, and more preset work, but I also want to get these improvements out there.

Let me know if there are any objections.

InsaneFirebat:
I'm in no hurry to get this stuff pushed out. I could probably pull more but I've been distracted with other things for a while. No objections either way.

idlechild:
Now just need to test this combined build a bit and then it's ready.

There was a graphical issue reported today, it's an old issue but it might have an easy fix to I plan to look into it later today. 
 The red in the background turns blue:
Load category preset "Lower Norfair > Ridley Farm Room" (PKRD)
Pause the game (Start), open the menu (Select, Start), unpause the game (start)

idlechild:
Also we got more discussion on the pause freeze issue.  Copying this from the discord:

if you mash the button to pause the game, open the menu and load a quick save in the wrong order you can cause a hardlock.

I commented that seems like it'd be easy to avoid and thus a lower priority.  That prompted azder to chime in:

i experience the pause menu hardlock very often, so i disagree that it's hard to reproduce
or maybe not that but rather that it's easy to avoid, unless i just don't understand how
tends to happen any time i'm doing a pause to do a menu that i'd legitimately be doing in an any% run

InsaneFirebat:
> The red in the background turns blue

That's caused by running one of the message box routines during the pause menu. It uses some of the same RAM as the pause menu. I haven't looked into that in quite a while, but it's also the reason I don't allow menu color customization while paused in my custom builds.

InsaneFirebat:
These shared memory addresses were the problem, and this init PPU routine (and its counterpart, restore PPU) is when it happens.

http://patrickjohnston.org/bank/85#f8143

$7E:3300..34FF: Backup of gameplay palettes during pause screen
$7E:33EA: Backup of HDMA channels to enable ($85) during message boxes
$7E:33EB: Backup of BG3 tilemap base address and size ($5B) during message boxes

Also added reserves to the status icons and made the health bomb flag smarter by changing color when it's ready to turn off. That should better illustrate the odd healthbomb flag behavior (only updating on drop spawns) to players. Didn't mean to drop more stuff on you at the last minute, but I forgot this was ready to be pushed from earlier this week.

PR #75, opened 2022-01-04
Add layout and cutscenes menu

idlechild:
All the same layout menu stuff as before, but now with a cutscenes submenu added to the game menu.  Currently has two intro cutscenes and also fast MB setting.  If you enable just the first intro cutscene then you get to keep the missiles at Ceres.

Initial support for MB2/MB3/Zebes Escape custom presets included.  It needs work (especially MB3 preset) but it's mostly functional.  Improvements welcome, especially since I'll be working on other stuff.  More room layout and cutscene options are planned.


src/presets.asm
```diff
@@ -200,14 +200,13 @@ preset_load_preset:
 
     LDX #$0000
   .loop_path
-    LDA $C1 : STA $7FF000,X
+    LDA $C1 : STA $7F0002,X
```
idlechild, line 203: All my work to save the $7F bank, and now I see this... ugh.
Remember that row of boxes Ruya found?  This could be the reason... although Ruya typically uses custom presets that don't use this logic, so maybe not?  Maybe he started with regular presets in his practice session.

Using $7F0002,X should be fine since we haven't loaded level data yet, and when we do it will overwrite this.

src/presets.asm
```diff
@@ -297,7 +294,6 @@ preset_start_gameplay:
     JSL $80835D  ; Disable NMI
     JSL $80985F  ; Disable horizontal and vertical timer interrupts
     JSL $82E76B  ; Load destination room CRE bitset, door/room/state headers, tiles
-    JSR $A12B    ; Play 14h frames of music
```
idlechild, line 293: I thought I tried removing this once and had problems, but maybe I made some other changes at the same time that were no good, or maybe the music handling is just better now.  I didn't have problems removing this.

src/presets.asm
```diff
@@ -318,10 +314,6 @@ if !FEATURE_PAL
 else
     JSL $A08A1E  ; Load enemies
 endif
-    JSL $82E071  ; Load room music
-    JSR $A12B    ; Play 14h frames of music
-    JSL $82E09B  ; Update music track index
-    JSL $82E113  ; RTL
```
idlechild, line 317: From what I can tell, we don't need to load room music because the preset already tells us what music to load and we already did it.

src/presets.asm
```diff
@@ -352,7 +344,6 @@ endif
     LDA #$0004 : STA $A7  ; Set optional next interrupt to Main gameplay
 
     JSL $80982A  ; Enable horizontal and vertical timer interrupts
-    JSR $A12B    ; Play 14h frames of music
```
idlechild, line 348: OMG we were doing this three times?  Why??  I thought we were supposed to be saving the frames.

I went back in github history, and sure enough back in 2019 we still had this logic where we play 14h frames of music in three places.
https://github.com/tewtal/sm_practice_hack/blob/b26b7aa2901787bd2c28304e07c14ef003321cf7/src/presets.asm

I'm very tempted to put out a 2.3.5 release just for this update.  It would be nice to get some feedback on it to make sure it doesn't break anything.

src/presets.asm
InsaneFirebat: This change causes some of my presets to load OoB. Double checked the last_preset pointer and it is correct.

I also encountered one crash going through a door after the sound freaked out and played a constant annoying tone, but I think that might have been this change rather than the music changes. Haven't seen it since swapping back to $7FF000.

src/presets.asm
InsaneFirebat: If I load a random preset a couple times very quickly, then go through a door, the screen won't fade in on the other side and it appears to have crashed. I can still open the menu or reload the preset, so it's not really crashed.

src/presets.asm
idlechild: I don't understand why changing $7FF000 to $7F0002 would break anything.  It's just your unpublished presets?  I wonder what the difference would be.

src/presets.asm
idlechild: Interesting.  On the one hand we could really dig into why this happens.  On the other hand we could just prevent users from loading multiple presets within 20 frames of each other, which is sort of the same thing as calling JSR $A12B except less annoying.  If it happens because we are doing too much DMA stuff in the background or something like that, preventing back-to-back preset loads would be a simple fix.

src/presets.asm
```diff
@@ -232,18 +231,15 @@ preset_to_memory:
...
 
   .one
     %a8()
     LDA ($00),Y : INY : STA [$C3]
     %a16()
-    INX #5
     BRA .loop
 
   .done
-    PLX
```
idlechild, line 240: I haven't tested the changes to preset_to_memory yet, but just on code inspection, I don't see why this method is messing with the X register.

src/presets.asm
InsaneFirebat: Tested this again and it's not causing me any problems now. I think I had one of them set to $7FF002 by mistake. That seems to give similarly bad results as I was seeing this morning. Sorry for false alarm

src/presets.asm
InsaneFirebat: I've now managed to crash the audio engine twice by spamming preset loads. We should probably do something to slow this down, but even without a fix, this is still a nice improvement.

idlechild:
I think it's working!  I decided not to rush out a 2.3.5 and instead lump this in with the 2.4.0 update.  However if you want to use it ahead of time, go ahead.

Todo:  Add option to turn this off, in case you patch in a large sprite set with SpriteSomething and it overwrites our tile graphics.  I also want to improve the music logic; I think I know why it can crash if you load presets too quickly.  I'm also looking at Kejardon's compression method which is a little more optimized compared to vanilla, although it probably won't make much difference.

InsaneFirebat:
Can you toss this into your pull sometime? I noticed we're not clearing the health bomb flag when loading presets. Mostly a non-issue but it can affect things if the preset is only one hit away from sub-50 energy.
`    STZ $0E1A    ; Clear health bomb flag`

Seems good so far. Loading a little faster than I expected.

idlechild:
> Can you toss this into your pull sometime? I noticed we're not clearing the health bomb flag when loading presets. Mostly a non-issue but it can affect things if the preset is only one hit away from sub-50 energy. ` STZ $0E1A ; Clear health bomb flag`

Will do.  Speaking of variables not being reset, if you load a preset while the screen is fading in or out for a door transition, the next door transition is abrupt.  Probably another variable that needs resetting.

Also 100% Late Croc Ridley Wasteland Up preset LUL

idlechild:
Also 100% Late Croc Maridia Cleanup 2 Shaktool preset has some bad scrolling

InsaneFirebat:
> if you load a preset while the screen is fading in or out for a door transition, the next door transition is abrupt.
Oh I'm totally making that an option if it is what I think it is. I love the "lights on" transitions :D 

> Also 100% Late Croc Maridia Cleanup 2 Shaktool preset has some bad scrolling
Not sure why I only applied half of the fix here, but it just needs `STA $7ECD21` added before the `LDA #$00` @ line 454.

> Also 100% Late Croc Ridley Wasteland Up preset LUL
I still don't know what causes this or how to fix it. There's no door ASM to correct scrolls, and the scrolls are set properly anyways. It even works properly with custom presets. I might try remaking it altogether.

InsaneFirebat:
I guess I'll open up a pull with these two preset fixes. New preset data fixed the Wasteland Up example. I'm gonna try to find more of those and try to fix them the same way.

idlechild:
Tile graphic optimizations complete.  Hopefully saved a few more frames off of that 30 frame example.  There isn't much more to save unless we store all of the raw level data.

Also finally done messing with the music.  The loadsaveoff option removed.  The normal music off option now does what we want it to; it tries to be as fast as possible when loading savestate or preset.  When turning music on, we assume the music data is no good and do the work to resend it to the SPC.  It should no longer be possible to crash when repeatedly loading random presets.

InsaneFirebat:
It takes a whole lot of spamming random presets, much more than before, but I am still able to crash the audio. I hadn't toggled music off before testing, so it was still at default. Wasn't able to crash at all with either music-off option. No big deal, just figured I'd let you know if was still possible.

idlechild:
That is both acceptable and disappointing at the same time.

Previously there was a loophole, if you got multiple presets in a row that all had the same data but different tracks, then you could fill up the music queue.  I'm not sure what would happen next, but probably at some point music data couldn't be queued, you'd play music with the wrong data and crash.  That loophole should be closed though.

As you said not a big deal, if that's the worst we can do then it is working pretty well.

idlechild:
I'm not sure why I made the preset music logic so complicated.  Loading music after a savestate is complicated, but for presets we should be able to clear the music queue and then that makes it relatively simple.

Here's the thing:  If I set there and spam loading random preset, it doesn't crash and it plays the correct music (if it crashes then it is rare).  However if I walk into the next room, there's a good chance the game will hang on the next room, even if the next room is playing the same music as the previous room.  It doesn't seem to happen if wait a bit between loading random presets, but it happens a fair amount if I start spamming.

Guess I need to review the vanilla music logic some more, argh really thought I was done this time.

InsaneFirebat:
I've done my worst. You've won this round.

idlechild:
Alright that's it for room layout.  I could work on more cutscenes but it's not worth holding up 2.4.0 release.

I'm planning to work on the MB2/MB3/Zebes Escape presets and prepare for release, could be soon (tomorrow?)  I would like to release this weekend.  If you've got any features, perhaps some controller menu options or infinite Ridley fights, please get them in!

idlechild:
It's true the extra vertical line on the menu isn't really noticeable, looks fine to me.

Just about ready to release.  I'd like to make the MB3 and Zebes Escape presets work a bit better.  There's more I'd like to do, for example Ceres Escape and Zebes Escape should be built-in presets.  However it's not worth holding up the release for it.

InsaneFirebat:
I don't think it's been mentioned anywhere but chabo added more presets to the Early Ice KPDR category. Worth adding to the changelog.

PR #76, opened 2022-01-10
[2.3.5] Scroll fixes for custom presets

InsaneFirebat:
This pull adds scroll data to custom presets. Since that data is wiped out by vanilla loading routines, the data is loaded later in the preset_scroll_fixes routine. I only used the MVN/MVP commands because I never have before. I don't actually know if it saves time vs a loop, but it seemed like an easy way to get around the Y register limitations (no LDA long,Y).

I went through all the hundo presets (which we don't have the original movie for) and found some places that would benefit from having the new layer 2 data. Also fixed a couple preset scroll bugs, one of which was mentioned in #75.

### DIFF
--- a/src/custompresets.asm
+++ b/src/custompresets.asm
@@ -8,7 +8,7 @@ custom_preset_save:
     LDA !sram_custom_preset_slot
     ASL : XBA : TAX ; multiply by 200h (slot offset)
     LDA #$5AFE : STA $703000,X   ; mark this slot as "SAFE" to load
-    LDA #$01B8 : STA $703002,X   ; record slot size for future compatibility
+    LDA #$01EA : STA $703002,X   ; record slot size for future compatibility
     LDA $078B : STA $703004,X    ; Elevator Index
     LDA $078D : STA $703006,X    ; DDB
     LDA $078F : STA $703008,X    ; DoorOut Index
@@ -67,6 +67,15 @@ custom_preset_save:
     LDA $0919 : STA $7031B2,X    ; Layer 2 Y position
     LDA $0921 : STA $7031B4,X    ; BG2 X offset
     LDA $0923 : STA $7031B6,X    ; BG2 Y offset
+
+    PHX : PHB
+    TXA : CLC : ADC #$31B8 : TAY ; Y = Destination
+    LDX #$CD20 : LDA #$0031      ; X = Source, A = Size
+    MVN $7E70                    ; srcBank, destBank
+    PLB : PLX
+
+    ; next available byte is $7031EA
+
     RTL
 }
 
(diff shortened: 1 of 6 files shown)


InsaneFirebat:
I'm getting ready to push my controller bindings menu and wondering how much you would mind if I replaced a bunch of two-byte vanilla addresses with labels. This is the list I use in my builds: https://pastebin.com/xtEatam5

I'll hold off on it until you give the OK @idlechild

src/custompresets.asm
```diff
@@ -132,14 +141,20 @@ custom_preset_load:
...
-    INX : LDA $703002,X : CMP #$01B0 : BEQ .done_loading
+    INX : LDA $703002,X : CMP #$01B0 : BEQ .done_no_scrolls
 
     LDA $7031B0,X : STA $0917    ; Layer 2 X position
     LDA $7031B2,X : STA $0919    ; Layer 2 Y position
     LDA $7031B4,X : STA $0921    ; BG2 X offset
     LDA $7031B6,X : STA $0923    ; BG2 Y offset
 
-  .done_loading
+    LDA $703002,X : CMP #$01BA : BMI .done_no_scrolls
```
idlechild, line 151: Should be #$01B8

src/presets/hundo_data.asm
```diff
@@ -2819,9 +2857,10 @@ preset_hundo_ridley_kihunter_shaft_up:
     dw #preset_hundo_ridley_wasteland_up ; Ridley: Wasteland Up
     dl $7E078D : db $02 : dw $9A3E ; DDB
     dl $7E079B : db $02 : dw $B5D5 ; MDB
-    dl $7E090F : db $02 : dw $0000 ; Screen subpixel X position.
     dl $7E0911 : db $02 : dw $0500 ; Screen X position in pixels
-    dl $7E0913 : db $02 : dw $EC01 ; Screen subpixel Y position
+    dl $7E0913 : db $02 : dw $EC02 ; Screen subpixel Y position
+    dl $7E0917 : db $02 : dw $03C0 ; Layer 2 X scroll in room in pixels?
+    dl $7E0921 : db $02 : dw $03C0 ; BG2 X scroll offset due to room transitions (Translates between Layer 2 scroll and BG2 scroll)
     dl $7E09C2 : db $02 : dw $03F3 ; Health
```
idlechild, line 2864: I think firefleas preset is also broken, I just noticed it last night

idlechild:
Did you want this to go out soon as a small update?  (without the controller bindings stuff, that'll be 2.4.0)

I'm fine with the RAM labels.

src/presets/hundo_data.asm
InsaneFirebat: I put a fix for that one in scrolls, line 434. Looks like I forgot to comment the room name. I'll add that in.

src/presets.asm
```diff
@@ -450,7 +462,8 @@ preset_scroll_fixes:
     LDA #$02
     STA $7ECD20 : STA $7ECD21
     BRA .done
-+   CPX #$D8C5 : BNE .done       ; Pants Room (door to Shaktool)
++   CPX #$D69A : BNE .done       ; Pants Room (door to Shaktool)
```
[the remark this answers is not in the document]
InsaneFirebat: I like that it keeps things in-line, and I don't have to name anything. I'm sure you've noticed how bad I can be at naming things :)

line 220 stood out pretty well.

InsaneFirebat:
I don't really mind when it goes out. If you wanna merge it sooner, I can just push some little fixes and spin up the next pull instead.

InsaneFirebat:
Ok, this pull can be considered complete (unless we still need to address that preset size comparison). Anything further will go to a new pull.

src/custompresets.asm
idlechild: For some reason I thought this was a BEQ and not BMI

idlechild:
I'm also good either way.  Leaning towards bundling it with the other updates.  Music fixes should be finished soon and then it's back to room layout changes (anti-softlock patches).

InsaneFirebat:
Added a fix for Botwoon RNG pointed out by stashiocat

PR #77, opened 2022-01-14
Controller bindings, RAM watch, and some things

InsaneFirebat:
### Controller Bindings Menu
This menu is far from perfect, but it does try to find a place for the input you're unassigning while still following SM's rules for rebinding. There is a case where it will fail to find a suitable input. I believe it occurs when assigning an angle and there already being an empty binding. It's easy for the user to fix and I got tired of reinventing the wheel on this one. If it's interesting to you, feel free to do whatever with it. It also comes with a menu for selecting common control schemes found on the wiki's spreadsheet. http://controls.spazer.link

### Common RAM Addresses Submenu
I added a submenu to the Customize RAM Watch menu for selecting common (or just useful to runners/testers) RAM addresses. The enemy properties were consolidated and given a selector for which side of RAM Watch to apply it to. The new common addresses will ask which side to apply to when selected, and then it jumps back to the main RAM Watch menu.

### Misc
In order to fit the new submenu in RAM Watch, I extended the menu background by an extra row at the bottom. Some extra options were added to the equipment menu for setting current ammo, like we have for energy/reserves. And finally, those labels for vanilla addresses are scattered all over the diffs.


InsaneFirebat:
Added in some docs that assume v2.4.0, some PAL equipment loadouts, an infinite Crystal Flash loadout, a fix for accidentally using lowercase in a header (it also breaks the footer), and a fix for my lowercase header.

idlechild:
Overall looks good.  I may make a few adjustments as I pull this in.  I'm not sure about making the menu taller, but I should test it out before passing judgment on it.

Also since I'm bringing in VARIA features like room layouts and fast MB, and VARIA allows angle up and angle down remapped, I'd like to allow that.  Preferably there'd be an extra setting you have to toggle on first, so by default it's like you have it here.

InsaneFirebat:
Oh, good thing I didn't make it wider too like my own builds xD Honestly, you probably wouldn't notice if you couldn't see the code. If you change your mind and decide you love it, going wider makes those choice list menus a lot easier to name ;)

Issue #62, opened 2021-09-29
Feature Request: Sub-menus for selecting possible Phantoon patterns


jberq:
Sub-menus for each Phantoon phase with checkboxes toggling the possible patterns which could appear would be useful both for players learning Phantoon as well as those optimizing and practicing the fight.

If a player learned, say, Fast Left 1 and then learned Fast Right 1, they could then practice Fast Left 1 _or_ Fast Right 1. The RNG wouldn't necessarily even need to resemble Phantoon's actual RNG— as long as they were reacting to different patterns it would be efficient and valuable practice.

InsaneFirebat:
The reason we don't offer more control over Phantoon patterns is because allows the rare patterns to appear at a much higher rate, even when these settings aren't used. It's been a while since I've looked into this, but I don't remember there being an easy way around it.

tewtal:
Yeah I don't think it's easily doable with just RNG value changes, but we'd probably have to add a separate branch to Phantoon's AI that does the pattern selection so we could implement locked patterns outside of the normal functionality. That way we can leave the original code intact as much as possible. Either way it's a pretty big project most likely, but an interesting one.

PR #79, opened 2022-01-17
Add toggles for individual Phantoon patterns (#62)

NobodyNada:
As specified in #62, allow the user to select a subset of Phantoon patterns to practice.

Implemented using hooks in Phantoon's AI that override the default RNG-based pattern chooser with a custom implementation that chooses patterns from the user-selected subset. If all the patterns are enabled (or all are disabled), the custom logic is disabled and Phantoon's regular RNG code is run instead -- so this feature should not impact
Phantoon's behavior or pattern probabilities when not in use.

Also includes an option to skip the long intro, because I got tired of sitting through it. With the "Skip Intro" option enabled, Phantoon begins the fight (starting from his wavy-fade-in animation) immediately when you enter the boss room.

### DIFF
--- a/src/defines.asm
+++ b/src/defines.asm
@@ -64,11 +64,11 @@
 !ram_room_has_set_rng = !WRAM_START+$5A
 !ram_kraid_rng = !WRAM_START+$5C
 !ram_crocomire_rng = !WRAM_START+$5E
-!ram_phantoon_rng_1 = !WRAM_START+$60
-!ram_phantoon_rng_2 = !WRAM_START+$62
-!ram_phantoon_rng_3 = !WRAM_START+$64
-!ram_phantoon_rng_4 = !WRAM_START+$66
-!ram_phantoon_rng_5 = !WRAM_START+$68
+!ram_phantoon_rng_1 = !WRAM_START+$60 ; round 1 pattern bitmask
+!ram_phantoon_rng_2 = !WRAM_START+$62 ; round 2 pattern bitmask
+!ram_phantoon_rng_3 = !WRAM_START+$64 ; eye close
+!ram_phantoon_rng_4 = !WRAM_START+$66 ; flames
+!ram_phantoon_rng_5 = !WRAM_START+$68 ; next flame
 !ram_botwoon_rng = !WRAM_START+$6A
 !ram_draygon_rng_left = !WRAM_START+$6C
 !ram_draygon_rng_right = !WRAM_START+$6E
@@ -401,6 +401,7 @@
 !CUTSCENE_SKIP_CERES_ARRIVAL = #$0002
 !CUTSCENE_SKIP_G4 = #$0080
 !CUTSCENE_FAST_MB = #$0100
+!CUTSCENE_FAST_PHANTOON = #$0200
 
 !COMPRESSED_GRAPHICS = #$0001
 !COMPRESSED_PALETTES = #$0002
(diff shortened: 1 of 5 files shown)


idlechild (commented):
A lot of work here, I'll need to try this out. I like the idea of fast phantoon entry, and it makes sense you might want to practice some random patterns without practicing all of them.

I'll need to take a closer look to ensure RNG was preserved by default.  It looks like you tried to do that.

NobodyNada:
I definitely tried to preserve RNG; if the user has not modified Phantoon's behavior it should be identical to the vanilla game unless I missed something silly. The RNG hooks simply execute a verbatim copy of the patched-out code (see the `.no_manip` branches) unless the user has selected a custom set of patterns, with no additional RNG calls or anything like that.

Also, even if the user has overridden Phantoon's behavior we should still preserve the number of RNG calls per frame (we always patch out one RNG call but make up for it with one call at the beginning of `choose_phantoon_pattern`).

InsaneFirebat:
I like the intro skip here. That's another win for players that rely on presets as their savestates. I would have liked to have seen the 1st/2nd round settings kept, but this is a pretty good trade-off.

Unfortunately, this doesn't preserve the odds for rare patterns unless all patterns are enabled. Both the mid-left and fast-right patterns of the second round are appearing at a significant rate when any pattern is turned off. I'm fighting Phantoon with 3-4 missiles per round and getting one or two rare patterns per fight, with just one pattern disabled. With both slow patterns disabled I'm seeing quite a lot of them.

idlechild:
Once you start toggling off certain patterns, I'm not sure it is practical to maintain the vanilla odds.  The user might not want the vanilla odds in that situation anyway.

With an entire submenu just for phantoon, perhaps it's possible to also have the 1st/2nd round options (maybe not, I haven't given it much thought so it might be super complicated).

Another possibility is we add a toggle to control vanilla rng vs more balanced rng, and then any time you toggle vanilla rng on it automatically toggles all patterns on.  This would make it more clear that turning off patterns changes the rng.  Also you could then toggle vanilla rng off and leave all patterns on if desired.

NobodyNada:
My reasoning was: the rare patterns are rare enough that the user could just toggle them off they didn’t want to practice them. I can find a way to restore the vanilla odds if needed, though it would make the pattern chooser quite a bit more complicated so I wasn’t sure if that was necessary or even desired.

Separate 1st/2nd round options isn’t complicated at all. The reason I didn’t implement it is mainly UX: it would have meant twice as many menu options to scroll through. But if that feature is valuable to people’s workflows, I can bring it back.

InsaneFirebat:
Having the 1st/2nd round options would let you disable the rare patterns. Without it, you'd be disabling two extra patterns for the first round just to get them out of the second.

Can we move that intro skip flag to SRAM and copy the menu pointer (`phan_skip_intro`) into the Game > Cutscenes menu?

NobodyNada:
That makes sense, I didn't think about that! Thanks, I'll make those changes when I get the time (it might be a few days as I've got a busy rest-of-the-week).

idlechild:
There's no rush :)

NobodyNada:
I've implemented those changes:
Screen Shot 2022-02-06 at 12 46 49 AM

The toggles are now separated by round, and the intro skip is moved to the cutscenes menu. As before, if the patterns for a round are not modified, we use vanilla RNG to preserve vanilla odds. I've also updated the documentation (but not the changelog).

idlechild:
Looks good.

I'll likely pull this into my candidate near the end of the week in preparation of 2.4.1 release, although it might not happen until the following week.  I've got some GT classic practice to do :)

PR #80, opened 2022-01-18
Crossbank support for preset data

idlechild:


src/defines.asm
```diff
@@ -405,6 +405,7 @@
 !COMPRESSED_GRAPHICS = #$0001
 !COMPRESSED_PALETTES = #$0002
 !COMPRESSED_PALETTES_8BIT = #$02
+!COMPRESSED_TABLES = #$0004
```
idlechild, line 408: Freed up enough space that we can use raw tile tables too.  Just about all of PJBoy's uncompressed data is being used for preset loading.  The tile tables don't make that much difference though; if we get low on space we could remove this.

src/presets.asm
```diff
@@ -198,58 +182,119 @@ preset_load_preset:
     RTL
 }
 
-preset_to_memory:
+category_preset_load:
```
idlechild, line 185: There's a lot in this method.  I still need to run more tests.  I have verified that crossbank works, but there are three possible scenarios for when the crossbank happens and I'm not sure I've tested all three yet.

I did simplify the preset data format.  Previously it was dl $7Exxxx, db $02, dw $xxxx.  It was always $7E, and $7E is the only value that makes sense.  There was support for db $01 but we never used it.  By reducing the format to dw $xxxx, $xxxx, of course this saves a lot of space, but also it makes everything in the preset data a word.  That makes crossbank support more practical to implement.

I did make an assumption:  Once we have crossed the bank once, we won't again.  This relies on the preset data being less than one bank (we aren't even close to that limit) and it relies on the preset data being laid out in a sane order in the asm file.  If using the lua script I'd expect a sane order, but if doing some work by hand I guess it could be an issue?  I'd prefer the preset data be in order anyway.  It's okay to have gaps (e.g. hi jump preset's parent is kraid instead of bootless preset), but it needs to be in order.

src/presets.asm
```diff
@@ -568,65 +613,50 @@ org $809AC9
...
-org $EE8000
-  incsrc presets/kpdr21_data.asm
-  incsrc presets/rbo_data.asm
-  print pc, " preset_data bankEE end"
-
-org $ED8000
-  incsrc presets/gtclassic_data.asm
+org $EAE000
+check bankcross off
+print pc, " preset data crossbank start"
```
idlechild, line 618: Previously presets started at bank E8.  Now starting near the end of bank EA (after the tile tables which start in bank E8).  Currently the data ends halfway through bank EE leaving bank EF open.  I just added 86K of data and saved a bank instead of using three.

Granted space isn't a big issue on vanilla practice rom yet; if it is we can gut the tile tables.  The best feature here is the lower maintenance.  Simply list all of the preset data together, no need to worry about the individual sizes.

src/presets/100early_data.asm
```diff
@@ -1,4102 +1,3562 @@
...
+    dw $D8B2, $0000  ; Doors
+    dw $D8B4, $0000  ; Doors
+    dw $D8B6, $0000  ; Doors
+    dw $D8B8, $0000  ; Doors
+    dw $D8BA, $0000  ; Doors
+    dw $D8BC, $0000  ; Doors
+    dw $D8BE, $0000  ; Doors
+    dw $D8C0, $0000  ; Doors
+    dw $D8C2, $0000  ; Doors
+    dw $D8C4, $0000  ; Doors
```
idlechild: I wrote a C# program to do all of the preset data work for me.  It's not pretty but it worked.

tools/sm_teledump.lua
```diff
@@ -1,20 +1,19 @@
...
     { 0x078D, 0x2, 'DDB' },
-    { 0x078F, 0x2, 'DoorOut Index' },
+    -- { 0x078F, 0x2, 'DoorOut Index' },
     { 0x079B, 0x2, 'MDB' },
-    { 0x079F, 0x2, 'Region' },
-
+    -- { 0x079F, 0x2, 'Region' },
+    -- { 0x07C3, 0x6, 'GFX Pointers' },
+    { 0x07F3, 0x2, 'Music Bank' },
+    { 0x07F5, 0x2, 'Music Track' },
```
idlechild, line 16: I put these in address order, and I commented out stuff we don't need in presets, which was a surprising amount.

tools/sm_teledump.lua
```diff
@@ -73,12 +70,16 @@ local MEMTRACK = { -- {{{
     { 0x0A68, 0x2, 'Flash suit' },
     { 0x0A76, 0x2, 'Hyper beam' },
     { 0x0AF6, 0x2, 'Samus X' },
+    { 0x0AF8, 0x2, 'Samus subpixel X' },
     { 0x0AFA, 0x2, 'Samus Y' },
+    { 0x0AFC, 0x2, 'Samus subpixel Y' },
```
idlechild, line 75: I want to add subpixel support, so might as well have this here for the future.  It's not a big deal for most presets though; I'll probably have it default to 0000 or FFFF most of the time.  Currently it appears subpixels are set to 0000 anyway so it'll take a bit of work to support it... although interestingly if Samus' pose is running left or right then subpixels might be 3000 or D000; I guess the game gives Samus one frame of movement in that case.

tools/sm_teledump.lua
```diff
@@ -73,12 +70,16 @@ local MEMTRACK = { -- {{{
...
-    { 0xD7C0, 0x60, 'SRAM copy' }, -- Prob not doing much?
-    { 0xD820, 0x100, 'Events, Items, Doors' },
-    -- { 0xD840, 0x40, 'Items' },
-    -- { 0xD8B0, 0x40, 'Doors' },
+    -- { 0xD7C0, 0x60, 'SRAM copy' },
+    -- { 0xD820, 0x100, 'Events, Items, Doors' },
+    { 0xD820, 0x04, 'Events' },
+    { 0xD828, 0x08, 'Bosses' },
+    { 0xD870, 0x14, 'Items' },
+    { 0xD8B0, 0x16, 'Doors' },
```
idlechild, line 82: Yep we collected 96 bytes of SRAM for no reason, and also 256 bytes of "Events, Items, Doors" when we only needed 54 bytes.

src/presets.asm
```diff
@@ -568,65 +613,50 @@ org $809AC9
...
+print pc, " preset menu bankF2 start"
+  incsrc presets/gtmax_menu.asm
+  incsrc presets/hundo_menu.asm
+  incsrc presets/kpdr21_menu.asm
+  incsrc presets/kpdr25_menu.asm
+  incsrc presets/nintendopower_menu.asm
+  incsrc presets/pkrd_menu.asm
+  incsrc presets/prkd_menu.asm
+  incsrc presets/rbo_menu.asm
+print pc, " preset menu bankF2 end"
```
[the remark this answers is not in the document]
InsaneFirebat: I believe @UNHchabo attempted to split the preset data into separate banks before we had the extra MB added. They should be safe to delete. Archived in the commit history at least.

src/presets.asm
InsaneFirebat: Are you aware of how GT Classic is setup with both boots and bootless routes together? Is that the issue you mention here? I believe I did the same thing for LN presets in PRKD. There's a separate group for No-X presets up to Ridley.

tools/sm_teledump.lua
```diff
@@ -2625,10 +2626,13 @@ local function save_preset(step)
...
-            preset_output = preset_output ..  "db " ..  tohex(size, 2) .. " : "
-            preset_output = preset_output .. (size == 1 and "db " or "dw ") ..  tohex(val, size == 1 and 2 or 4)
-            preset_output = preset_output .. " ; " .. annotate_address(addr, val) .. "\n"
+            preset_output = preset_output ..  "    dw " ..  tohex(addr - 0x7E0000, 4)
+            -- preset_output = preset_output ..  "    dl " ..  tohex(addr, 6) .. " : "
+            -- preset_output = preset_output ..  "db " ..  tohex(size, 2) .. " : "
+            -- preset_output = preset_output .. (size == 1 and "db " or "dw ")
+            preset_output = preset_output .. (size == 1 and ", $??" or ", $")
+            preset_output = preset_output .. tohex(val, size == 1 and 2 or 4)
+            preset_output = preset_output .. "  ; " .. annotate_address(addr, val) .. "\n"
```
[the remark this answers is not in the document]
InsaneFirebat: Haven't tested the data it produces but the script runs without any obvious issues.

src/presets.asm
idlechild: Yep that's what I was referring to.  It seems like they still work, and I do see the preset data is laid out in a good order so it should work.  A bit of a scare seeing ice/spazer not selected before ridley in the No-X presets... but then I went back to 2.4.0 and same thing so I think there's just an issue with the preset data.

idlechild: I was going to ask, do you have any more category presets, or are they all uploaded at this point?

InsaneFirebat: Interesting. I didn't realize that mistake ever made it into the main branch. I planned to remake most of the presets again when the MB preset stuff is finalized, but that's pretty easy to fix anytime.

InsaneFirebat: I have KPDR 22% (adds Kraid tank) and PRKD 19% (skips wave missile). There's also a couple NewGame+ routes that I never planned to submit. 

Feel free to delete the All Bosses routes. I made those in a hurry for a multicategory tournament that featured the category and no one ever ran it outside of those races, much less used the presets. They're low quality and mostly wasted space.

tools/sm_teledump.lua
```diff
@@ -64,7 +61,7 @@ local MEMTRACK = { -- {{{
     { 0x09CC, 0x2, 'Max supers' },
     { 0x09CE, 0x2, 'Pbs' },
     { 0x09D0, 0x2, 'Max pbs' },
-    -- { 0x09D2, 0x2, 'Current selected weapon' },
+    { 0x09D2, 0x2, 'Currently selected item' },
```
InsaneFirebat, line 64: I would leave this commented out. It's very common to leave missiles or supers selected when setting up for a preset, and that's generally unwanted because the door will be blue when the preset is loaded. It's easier to add in lines to select power bombs than to delete a bunch of lines to turn off missiles.

src/presets.asm
InsaneFirebat: Might wanna throw a warnpc on this to keep it out of bank $F0 (infohud.asm). I just tested a build with all of my presets at once and ended up going over. Had to move the tile tables back a bank.

tools/sm_teledump.lua
InsaneFirebat: $7ED91A was cut from the events/items/doors data, but it appears to have 52 instances in my any% presets. RAM map says it's the "global number of items loaded counter". I could only find a couple instances of it being incremented in bank $84, but not sure if it's ever used for anything. $7ED90C came up as well, and probably handles the Tourian map which is marked blue when you first enter iirc. That's probably not useful but figured I'd point it out in case it was overlooked.

src/presets.asm
idlechild: There is a warnpc $F08000, it didn't fail the build?

tools/sm_teledump.lua
idlechild: I saw $7ED908-0F was map stations and I thought that wouldn't be relevant for presets... although I suppose maybe it would be for a true completion run or something.  I forgot Tourian is auto-filled in.  Yep we should have those values in the presets.  I expect $7ED91A is useless.

tools/sm_teledump.lua
idlechild: I've been thinking about saving/loading which doors are open.  I don't believe it is directly tracked by the game so not sure how feasible it is.

Even if it can be done, I'm not sure what's going to happen when you run the lua scripts.  Does it play through with the vanilla rom?  The practice rom that they were recorded with?  Or can the latest practice rom be used with the previously recorded inputs?  That'd be the only way door tracking could be added to the lua scripts and have it actually work.

Also if you have any head start on door open stuff that might be useful.  iirc you looked into once a while back.

src/presets.asm
InsaneFirebat: I think one of them isn't even the correct route. Maybe some lua mistakes that caused a few broken presets. It's been a while and I've had zero motivation to look at them again.

tools/sm_teledump.lua
InsaneFirebat: I have no progress on that, but here's the chat I had with Smiley and PJ about it.
https://discord.com/channels/127475613073145858/370769325851803648/867761505872642048

InsaneFirebat:
The emulator won't let you play back a movie from a different ROM than it was made with.

tools/sm_teledump.lua
InsaneFirebat: The emulator won't let you play back a movie from a different ROM than it was made with.

tools/sm_teledump.lua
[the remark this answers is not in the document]
InsaneFirebat: I have tested it now and all is well.

src/presets.asm
InsaneFirebat: I must have just missed it when applying the changes to my branch. Thought I looked for it before posting but I must not have since I replied here instead of adding a new inline comment.

tools/sm_teledump.lua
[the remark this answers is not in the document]
InsaneFirebat: Also, you only need the emulator that the lua was written for (idk why, but they're usually not compatible across emu versions). Here's the one we use for presets: https://github.com/TASVideos/snes9x-rr/releases/tag/snes9x-151-v7.1

My preset movies are here if you wanna try it out: https://github.com/InsaneFirebat/sm_practice_hack/tree/CustomROMBase/tools/movies

tools/sm_teledump.lua
idlechild: Oh duh, instead of a new variable determining which doors to open, I should go with your idea to make all of the doors open (or at least all of the blue doors).  That's simpler and no need to worry about lua.

My plan was to search the PLMs but I see from the discussion that's not going to work.  Searching the BTS for 40-43 would probably work though.  I'll look into that.

idlechild:
I was hopeful this logic would open all blue doors, but it doesn't work for doors that used to be non-blue.  Other than that it works reasonably well, and it makes practicing climb more fun (let me know if there are any other rooms like climb).

Also beware my crossbank preset work may have an issue.  When I'm loading random PRKD presets, there's at least one preset that loads a bunch of garbage graphics and softlocks the game.

InsaneFirebat:
While this is awesome for most of the presets, a few of them are designed to be able to practice opening the door as part of the next room setup. Terminator and Springball Maze, for example. Can we have an option to open doors in the presets menu?

idlechild:
Good idea.  Also there may be some people that don't like all the open doors.

I'm also planning to take another look at the blue doors.  If the preset has used a super or power bomb to open the door already, then I'd like to have the door automatically open instead of having the blue door there.

Also I need to fix climb room.  Looks like there is some special BTS values used for the half-tile platforms.

InsaneFirebat:
Oh, I hadn't seen that yet. I was just going to mention the Bat Cave presets. Doesn't seem to happen with all rooms that have vertical doors opened, but both of those rooms do.

idlechild:
@InsaneFirebat this keeps the firefleas false wall door closed, unless you activate the scroll before you save the preset

idlechild:
I just noticed something.  Enter wrecked ship basement in the center(ish) of the door, so Samus is centered when entering basement.  Save custom preset and load custom preset.  Notice how much the screen is shifted.  Sometimes you get a graphical glitch on the right side as well.

Why is the screen shifted so much?  What are we missing?

InsaneFirebat:
The distance Samus moved since last frame is causing the camera shift. Samus' last X position can be set to the current X position to fix it. `LDA $0AF6 : STA $0B10` Might wanna do it for Y position too.

After our preset code has run, these two routines calculate that value and then use it to update the screen position.
http://patrickjohnston.org/bank/90#f96C0
http://patrickjohnston.org/bank/90#f95A0

tools/sm_teledump.lua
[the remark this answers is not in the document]
InsaneFirebat: There is a bug here on line 2633. We're getting $$xxxx for the value. I tested by removing the second $ on that line, which worked, but I don't exactly know what the other one means.

idlechild:
Yes please!

InsaneFirebat:
Upon testing for background bugs, I found that the lower route after Ice Escape in KPDR21 still needs layer 2 data. I'd throw this into my pull, but my pull doesn't have the new preset format yet. Perhaps you can bring it over to your pull?
https://github.com/InsaneFirebat/sm_practice_hack/commit/b6b2755e6c87a191813317703bc11a3eb11fed87#diff-89c19d7e555d4ab01d60beef07bcf7d7ae37d8928c6c1417c5ca5521217b8836R1954

idlechild:
A little unexpected considering I saved the presets while Samus wasn't moving at all, so not sure why it is calculating that Samus moved... but whatever I'll take it.  Good find.

InsaneFirebat:
Same here. Unless it was lazily added in later, all of my presets were made from a standing, crouched, or morphed position while not in motion. I'm even more confused at how it still thinks that after reloading the same preset. I would have expected the values to be sane after the first load.

idlechild:
Thanks IFB, presets really starting to look good now.

I was able to get a 59'87" with the Ceres Escape preset so I'm ready to add those.  Thanks for the tip on 50.B000 start.  I also found out I need to arm pump just as you enter the door.

I see both Zeni and Ruya do an extra arm pump, so their setup must put them in the 51.B000-51.DFFF range.  I'll make the built-in preset use 51.B000.

idlechild:
I plan to copy-paste the first three late croc presets to all other preset categories.

It'd be nice if the landing site background was correct but it's not a big deal.

InsaneFirebat:
I found a few more presets that stand on blue doors. I'm going to add another way to bypass the door opening on my branch, but on a per preset basis. I'm going to manually add something like `    dw !PRESET_DOORS, $0001 ; doors closed` and `    dw !PRESET_DOORS, $0000 ; doors opened` to specific presets and check that after the !sram_preset_options variable. If you wanna do the same, I can pull all those preset changes for v2.4.2, or you can grab them from one of my branches and merge them whenever.

idlechild:
When testing, I found I wanted the doors closed for the pit room revisit since Samus just got off the elevator and needs to open the door.  Of course!  Instead of checking if Samus is just getting off the ship, I should be checking if Samus is T-posing.  T-pose = doors closed.

T-posing (pose = 0) can also be used to keep Samus standing on top of a closed door.

src/presets.asm
```diff
@@ -753,7 +753,7 @@ preset_scroll_fixes:
     LDA !sram_custom_preset_slot
     ASL : XBA
     CLC : ADC #$31E9 : TAX       ; X = Source
-    LDY #$CD52 : LDA #$0031      ; Y = Destination, A = Size-1
+    LDY #$CD51 : LDA #$0031      ; Y = Destination, A = Size-1
```
idlechild, line 756: An easy mistake to overlook, but I'm surprised we hadn't noticed this until now.  It was apparent something was wrong when I saved and loaded a custom preset at Pancakes and Wavers.  Just outside Wrecked Ship at East Ocean was also not good.

Fortunately the mistake was on load and the scroll data that was saved is correct.

In other news scrolling sky room backgrounds look good when loading presets now.  I tried turning scrolling sky back on but same issue as before, so leaving scrolling sky off.  It looks fine as is.

idlechild:
I just couldn't let it go :)

I'm not sure exactly why, but deferring the scrolling sky room setup asm prevents the pancakes and wavers corruption.  Perhaps it requires all the interrupts to be enabled first.  The other changes make sure all the necessary VRAM is initialized per Samus' location.  It's rather specific to the vanilla game so I included that logic with the tilegraphics stuff (also saves some space in bank 82).

InsaneFirebat:
I was pretty happy with the last commit fixing all the broken spots. Nice work getting it moving again.

InsaneFirebat:
This Climb Down preset has crash potential if you hold right while it loads, much like the main street preset we were looking at on my stream some days ago. Added a commit to my pull that updates the previous pose and fixes the crash.

idlechild:
Good find.  Also your named defines came in handy.  I went to put in a define for previous pose and found pose direction/previous pose direction.  We should probably set that too.

idlechild:
Lol I should also build before pushing, just a sec so I can fix that...

InsaneFirebat:
The PAL build has a crash with the ship landing sequence when using the preset. 

There's another issue with pausing in PAL that's been there for a while. I tried commenting out `ih_hud_code_paused` and it's still not working. Code and addresses looks the same for both regions, but the game just won't pause with the Start button. The map station opens it successfully, but the issue persists after unpausing.

InsaneFirebat:
The PAL pause issue is fixed on my pull. The ship landing sequence just needed slightly different pointers for locking Samus. Seems the unlock pointers are the same as NTSC. Fix below.

```if !FEATURE_PAL
    LDA #$E8CA : STA $0A42 ; Lock Samus
    LDA #$E8D9 : STA $0A44 ; Lock Samus
else
    LDA #$E8CD : STA $0A42 ; Lock Samus
    LDA #$E8DC : STA $0A44 ; Lock Samus
endif```

**Comment by idlechild** [comment]
Well that's what I get for making assumptions.  I should have looked up where these routines were.

**Comment by idlechild** [comment]
I always think I'm saving space in bank 82 and then it ends up being bank 80.  Oh well probably good to save space there too.

On the one hand, the scroll preset fixes aren't used by custom presets anymore thanks to the scrolls being saved with the presets.  On the other hand, the category preset fixes are a lot of customized logic... for presets... so it kinda fits in the custompresets file.  I'm expecting we'll make more category preset fixes in the future and I think this will keep the regular presets file from getting cluttered.

File: src/presets.asm [hunk 12]
```diff
@@ -296,10 +405,19 @@ if !RAW_TILE_GRAPHICS
 else
     JSL $82E7D3  ; Load level data, CRE, tile table, scroll data, create PLMs and execute door ASM and room setup ASM
 endif
+    JSL preset_scroll_fixes
+
+    LDA !sram_preset_options : BIT !PRESETS_CLOSE_BLUE_DOORS : BNE .done_opening_doors
+    LDA !SAMUS_POSE : BEQ .done_opening_doors
```
InsaneFirebat, line 411: This only covers presets without any suits. $009B is the pose for facing forward with Varia or Gravity.

InsaneFirebat:
That dev feature came just in time. I just made myself some new any% presets and now it's time to test them. It seems to skip over the first preset in the group but goes back to it before starting the next group.

idlechild:
> That dev feature came just in time. I just made myself some new any% presets and now it's time to test them. It seems to skip over the first preset in the group but goes back to it before starting the next group. Edit: It also skips the second preset of each group entirely.

It skips the second preset entirely? :(
I was aware it temporarily skips the first preset of each group and then comes back to it; it was just easier to code it that way. 
 Not good if it is skipping presets though.

idlechild:
New shinetune feature added.  I still need to clean it up (literally, I think the numbers will just stay there if you switch infohud modes) but it's there if people want it.  dagit was looking for some feedback on multi-stutter short charges.

It's average momentum over first 44 frames, although the average is skewed slightly (2.4% or factor of 1.024) to make the math simpler.  The average is slightly higher than it should be.

idlechild:
Alright the presets should be loaded in order now if you turn the rng off.

InsaneFirebat:
Can we keep the Last 3 Rooms presets? If you don't care about the timer, this one is a lot more useful than the escape start.

Also, how are you generating these? Are they manual edits or are you using the lua?

idlechild:
lol I suppose, I wasn't expecting anyone to vouch for the last 3 room presets, but sure I'll bring them back, we have plenty of space for presets.

I used a tool to change the preset data format.  In fact I was planning to run the tool again at some point to remove BG2 offsets since I think those are only needed for Ceres and can be set to 0 otherwise, and our preset scroll fixes already have the BG2 offsets for Ceres.  I think we need layer 2 X/Y but not BG2 X/Y offsets.  Anyway if your new KPDR presets are in the old format, that's no problem I can convert them to the new format.

Aside from the format updates, the other changes I've been doing manually.  So far it's mostly been copy-paste from other presets.  Occasionally I save a custom preset and then pull values out of there to put into the category presets.

idlechild:
Music needs a little work; should probably set the presets to start with no music.  Also wrong music plays for MB3 preset.  Also should clean up the MB graphics for Zebes Escape.  Otherwise MB2/MB3/Zebes Escape presets are functional, I just need to add them to other categories.  Bringing back last 3 room Ceres presets also on my todo list.

NobodyNada:
[`REP count`](https://rpghacker.github.io/asar/manual/#rep) is an ASAR pseudo-instruction that causes the following instruction to be duplicated, which I used to make NOP filler bytes to pad the hooks to the correct length. Not to be confused with the `REP #imm` instruction for clearing status flags. Yes, that's very confusing and yes, I should have added a comment (but didn't think about that at the time). BRA is clearer and faster anyway.

Good catch on the inverted patterns. I tried to preserve vanilla behavior, but since I use emulator savestates I never encountered or even thought of practice ROM savestates affecting the frame counter. IMO it would make the most sense to keep the menu just how you implemented it, except for having the inverted patterns default to off unless they are explicitly turned on. I don't imagine many runners need or want to practice them, so if they default to vanilla they'll probably be an annoyance (due to savestates) more than a benefit.

idlechild:
Thanks for the explanation; I was not familiar with that form of rep and definitely broke it when merging.

Inverted setting now defaults to either vanilla or off whenever you change any other phase 2 setting.

InsaneFirebat:
I finally got around to testing this out since the 2nd round choices were added. Everything is working great. Nice work!

I remember seeing the rep command in asar's docs and thought it was an odd choice to use the same name as an opcode. I think I'd rather write out repeat than have a confusing rep.

idlechild:
I added the specialized fixes but without the preset fixes flag.  If conditions are right then do it!  Also added one for parlor.  Also fixed an oversight where custom presets were not getting the ceres BG2 X/Y offset fixes.

I moved custompresets to another bank (83) since the menu bank (B8) is almost 75% full and there isn't a good solution if it actually becomes full.

InsaneFirebat:
I've been looking into the menu freespace issue since it already filled on my branch with all the customization stuff. There's not a whole lot that needs to be done. Mostly just common routines that need to be converted to JSL's all over the place, and some fixes for assumed DB's. Total did most of the work toward it when he allowed preset_menus to live in different banks.

InsaneFirebat:
I went ahead and put this branch together as an example. The most drastic change is to the JSRs attached to each menu item. I setup some bridges between the two banks to handle that. I also set the DB before each of those JSRs, which might be overkill. I think only the ones that use tables will need it. It needs more thorough testing but I did poke everything at least once to make sure there were no crashes. With all these changes (and before custom presets are moved), the MainMenu bank is at about 3800h bytes.

https://github.com/InsaneFirebat/sm_practice_hack/commit/4848aabcfae72a6cc1c75a7fb77351ed3d7928ef

src/custompresets.asm
```diff
@@ -466,6 +466,15 @@ preset_scroll_fixes:
     ; -----------------------------------------------
     ; Specialized Fixes (Category and Custom Presets)
     ; -----------------------------------------------
+  .specialized_parlor
+    LDY !SAMUS_Y : CPY #$00D0    ; no fix if Ypos > 208
+    BPL .specialdone
+    LDY !SAMUS_X : CPY #$0175    ; no fix if Xpos > 373
+    BPL .specialdone
+    %a16() : LDA #$00FF
+    STA $7F05C0 : STA $7F05C2
```
InsaneFirebat, line 475: Some of the later presets that don't have screw attack would benefit from all 6 tiles being removed.
    STA $7F0480 : STA $7F0482
    STA $7F0520 : STA $7F0522

idlechild: Yeah I would like to, not sure if I'm going to get to it.

I want to release this update soon, but feels like half of what I worked on is only half-done, still more work to do.

idlechild:
Theoretically done working on presets... although I should tweak the specialized logic to remove more parlor blocks as Firebat suggested.

I suspect I've made some copy paste errors :(
I may update my tool to try and catch errors.

idlechild:
Also still need to polish up some of the new features for this PR before it's ready to go.  I wanted to release it this weekend, but it might not be ready until early next week.

idlechild:
I knew updating the tool would be worth it, found some copy-paste errors.

Ideally the tool would be smarter and also work for menus, but this will do for now.  Copy-paste errors in the menu are a little easier to spot and also aren't as disastrous (may break one preset but not all presets after that one).

idlechild:
Finally done with preset data for this release (or at least I'm done working on the preset data)

InsaneFirebat:
I've been holding off on the preset_data changes so I only had to do it once. Will start merging and testing them now.

InsaneFirebat:
I've gone through almost all of the presets by now. The biggest issue is that the escape climb preset is a little broken and spawns you on the level above. Some of the categories that only had presets added to early game now have minor background issues. pkrd, 14ice, 14speed, and nintendopower all have examples of this starting in Brinstar. Most of those issues are somehow confined to Brinstar rooms. RBO's Brinstar presets are good, but some of the Norfair presets have bad backgrounds.

I'm not sure any of this should be considered a blocker. I think it would be easiest to fix these if we update teledump with all the new presets and run it through those preset movies again. I'd just rather wait until everything else in this pull goes through before I touch anything else with preset_data.

idlechild:
Thanks for the feedback!  It is not quick to check all of these presets.  Agree it's not blocking but will look into this, at least the escape climb preset.  I must have copied from a bad example.

I'm hoping to wrap up this release today or tomorrow.  I'll be posting an update for vanilla infohud soon.  I still need to adjust CPU usage (likely remove some NOPs) to account for the new logic added.

idlechild:
Alright I have "scored" the CPU updates and reduced artificial lag to compensate.  Unfortunately it's probably a bit more lag when HP updates, hopefully it's not noticeable in hell runs.  I had to reduce it quite a bit when vanilla infohud is turned on.

With that done, I just need to adjust the new shinetune average momentum feature, review the presets Firebat mentioned, potentially improve MB3 and Zebes Escape presets (hopefully at least fix the music for MB3) and then I think it's ready.  Probably won't be done tonight, so hopefully release tomorrow.

Any issues with this pull request?  Please let me know!

idlechild:
It's ready.  I'll be pushing the PR in a few hours.

InsaneFirebat:
14% Speed is missing the new Tourian presets in `_menu.asm`. All looks good otherwise.

PR #82, opened 2022-01-23
[2.4.1] Minor Fixes when Loading Presets

InsaneFirebat:
Clearing the timer at $7EC400 fixes the issue with doors fading out too fast (or not at all) after a preset was loaded during a door transition. Clearing $0797 fixes the inability to pause until using a door transition. Also clearing the health bomb flag for good measure.


InsaneFirebat:
Made a little force push to cleanup the other place we were clearing these flags. It wasn't working anyway because DB was for the preset_data and not hitting $7E.

InsaneFirebat:
I copied in the `STZ $1C1F` from the teleport code. That's the one that was giving us trouble in the `maybe_trigger_pause` routine, although I don't remember what that trouble was. Clearing `!ram_transition_flag` fixes the missing timers and/or excessively large transition time from loading a preset while in a transition. 

Moving the RTS in `cm_loop` saves a few bytes and fixes the corrupt water layer after leaving the menu with select or start buttons. It just needed to wait an extra frame before exiting the loop.

src/presets.asm
```diff
@@ -549,7 +550,7 @@ add_grapple_and_xray_to_hud:
 }
 
 print pc, " preset_start_gameplay end"
-warnpc $80FC00
+warnpc $80F800 ; save.asm
```
idlechild, line 553: Disappointed I missed this, good catch

InsaneFirebat: I have some extra stuff in there for randomizing equipment, which comes with $160 bytes of tables. I also moved the LoadRandomPreset routine there since I ran out of menu space and had to cut/move some stuff out of $B8. So I didn't catch it, someone getting a savestate build did :)

InsaneFirebat:
I came across some more bad scroll situations while going through all the doors. Added scroll fixes for all of them. 

I also came across a Wasteland preset that was morphed and realized it was stuck in place. There's a scroll tile at that exact position. Decided to track down why that causes problems and found that a table pointer wasn't being set. This pointer is for Samus' X speed and is cleared somewhere in the beginning of preset load. In almost all cases, it's set to $9F55 every frame. There are two other pointers for suitless liquid physics, but they're only used while Samus is in motion. Once the pointer is set to $9F55, the normal routine starts setting it every frame like normal, and the other pointers will be set if their conditions are met.

There was also a broken preset in GT Max% that I forgot to include the fix for. Apparently it's been broken from the beginning. Not sure how it would've happened either. I'll push that fix in just a few minutes.

src/presets/gtmax_data.asm
```diff
@@ -1476,24 +1476,25 @@ preset_gtmax_lower_norfair_wasteland:
...
+    dl $7E0A1E : db $02 : dw $1108 ; More position/state
+    dl $7E0AF6 : db $02 : dw $0246 ; Samus X
+    dl $7E0AFA : db $02 : dw $0489 ; Samus Y
     dl $7ED8BA : db $02 : dw $4301 ; Events, Items, Doors
     dw #$FFFF
 .after
 
 preset_gtmax_lower_norfair_metal_ninja_pirates:
     dw #preset_gtmax_lower_norfair_wasteland ; Lower Norfair: Wasteland
+    dl $7E078D : db $02 : dw $99EA ; DDB
```
idlechild, line 1497: Are you sure about this DDB?  Trying to figure out it's the same as the one above even though you are in a different room

InsaneFirebat: I wish I could tell you what's going on there but I don't have a clue. I had to remake those presets and copy in the data to see what was wrong in the diff. It has been tested though, and it works.

InsaneFirebat: It was correct for this preset but needed to change for the one above. For some reason, that line didn't transfer over when I merged from my other branch. Looks like everything else is there. Fixed in the next commit.

src/misc.asm
```diff
@@ -102,8 +102,8 @@ if !FEATURE_PAL
 org $90EA38
 else
 org $90EA3B
-    BRA $0B
 endif
+    BRA $0B
```
idlechild, line 106: Yikes!  How long did it take you to find this?

If the game is crashing when completing Ceres, I wonder why this would fix it.

InsaneFirebat: This change only fixes the pause check. I thought GitHub was playing tricks on me for a second there, but it didn't merge the Ceres elevator change because it was only missing on my branch. You must have fixed it already back when you were going over PAL differences.

InsaneFirebat:
Added a fix for opening the menu during the death animation, which causes the palettes to darken and the white text to disappear.

src/presets.asm
```diff
@@ -273,6 +273,9 @@ preset_start_gameplay:
     JSL $878016  ; Clear animated tile objects
     JSL $88829E  ; Wait until the end of a v-blank and clear (H)DMA enable flags
 
+    ; Set Samus last pose same as current pose
+    LDA !SAMUS_POSE : STA $0A20
+
```
InsaneFirebat, line 278: This causes a problem for presets that were saved while sparking. Before this change, Samus would start moving slowly and build up speed until crashing the spark normally. Now, it just traps Samus in the air with the sparking pose. I'm not sure how to handle it yet. Could get a bit messy to check for each spark pose, but I guess that's the last resort.

Unrelated to that, I just noticed that custom presets crash if saved during the spark windup pose.

InsaneFirebat:
I've added in the special preset fixes to remove power bomb and shot blocks here and there. The presets will need the changes from pull #80 in order to load correctly since I added them in the new format. So far I'm clearing blocks for Mission Impossible, Retro Brinstar Escape, Leaving Hi-Jump, and Firefleas. Also added a scroll fix for entering Kraid's lair, and included the changes to that routine from pull #80.

Kraid intro skip hijacks the LDA #const for the wait timer before the earthquake starts. If enabled, it loads one and returns. If disabled, load the normal value and return.

The Charge Timer HUD mode now counts up the number of frames that shot was released between charges and prints it to the left of the normal timer. It will only count up to 36 frames before clearing itself and resetting the charge timer to 61 frames. I chose 36 because it's the largest single-tile number that can be displayed when the minimap is enabled.

InsaneFirebat:
Merged with pull 80

PR #84, opened 2022-03-07
Menu Freespace Expansion and Crash Handler

InsaneFirebat:
The first commit separates the menu framework code from the mainmenu data/code. It also allows each of the menu categories on the MainMenu to live in separate banks. Any bank in the rom can be used as long attention is given to word addressing in banks beyond $BF. This is achieved by adding a table for MainMenu that lists the bank of each menu category. The new cm_mainmenu macro is used to know when to change banks, and a new menu action type was added for cm_submenu so that routines such as action_submenu can be addressed from anywhere. Finally, all of the JSRs in the menu's execute routines are converted to indirect JMLs with a return address, so nearly every mainmenu routine needs to return with RTL instead of RTS. References to menu JSRs have also been converted (in comments and labels).

The next commit is a crash handler that grabs register data and the top 30h bytes of the stack and saves it to SRAM. The menu is then opened, and the flag in SRAM is detected to jump to the new crashdump viewer where the data can be displayed. If the crash was triggered by a BRK instruction, then the data that it pushed to the stack (status register and long return address) will be duplicated above the stackdump and reformatted for easier reading. The crashdump viewer drops into a loop to monitor for input. In this first iteration, the user can only load a savestate, soft reset, or cycle through a few hardcoded palettes in case text readability is low. We may want to setup extra pages that can be navigated through to show additional memory addresses of interest.

Finally, there's a commit to pull Samus out of the acid when loading Zebes Escape presets. It changes Samus' Y position and puts her in the facing forward pose so that she hovers just above the acid. The player can press left or right to regain control when the acid is lowered (or immediately if they don't care) and still have time to wait for the escape text.


InsaneFirebat (commented):
Cleaned up and fixed a few things upon review. I'm happy with it now if it were to go through immediately. That said, I left it as a draft because I expect there might be more to play with in the crash handler. I've got a standalone version of it linked below that has support for multiple pages and doesn't rely on the menu. The extra pages could draw out some memory values we suspect to be problematic, or we could build an entire memory viewer into it. Any ideas?

https://github.com/InsaneFirebat/SM_Crash_Dump/blob/main/CrashHandler.asm#L162

InsaneFirebat:
The crash handler now runs without launching the practice menu. However, much of it has been moved into the menu bank (now $89) in order to easily share a few VRAM transfer routines. I went ahead and added a couple more pages. The first is the memory viewer mentioned above. It lets you lookup any address, RAM or ROM, and instandly see the value held there. It also prints 16 bytes near the targetted address. For example, if you point to $7E0A44, all 16 bytes at $0A4x will be displayed. The other new page is just a bunch of text with simple instructions for how to report the crash.

InsaneFirebat:
I've added the `COPHandler` routine and set its type to 2. Otherwise, it's exactly the same as `BRKHandler`. When drawing the BRK/COP data, the operand that followed the instruction will also be drawn, in case we want to start using them for debugging. The palette byte in the text strings is always #$28, so I removed it and hardcoded the value in `crash_draw_text`. I also shortened a few variable names and reverted the SRAM version bump since `!CRASHDUMP` was moved to $7E.

InsaneFirebat:
The latest commit replaces a few of the vanilla routines in an effort to minimize changes to RAM after the crash. It also fixes all of the issues with unreadable crashdumps that I'm aware off, such as crashing at boot or in the game over menu. Holding left or right in the memory viewer will delay for a bit, then scroll through the numbers at a fairly quick pace. Holding Y still scrolls 4x faster.

InsaneFirebat:
There's a build error from renaming variables before pushing it out. I'll clean it up in the next commit, which will likely be color palettes that make more sense.

InsaneFirebat:
Palettes and renaming fixes are in. A bunch of small changes to handle stack underflow/overflow are right behind it. I shifted almost everything up a bit in the tilemap to fit in that final line of the stackdump, and added in a warning message for the stack pointer. The stack and DP need to be repaired, and some of the data can no longer be trusted, as the stack was dead when that data was passed through it. The stackdump also targets the 30h bytes closest to the broken pointer.

InsaneFirebat:
I think I'm done with this. There isn't anything here for the players so it stay in limbo until then.

InsaneFirebat:
Added Pacifist Mode to Game options by request. If enabled, skip past damage calculation. Separate hijacks for shinesparks and power bombs. Everything else seems to run through the same routine.

InsaneFirebat:
Merged with https://github.com/tewtal/sm_practice_hack/pull/86

PR #86, opened 2022-03-15
Added ability to monitor spacetime beam parameters

idlechild:
Buyer beware:  I just made this, I made mistakes while doing it, which I fixed, but there may be more mistakes.  Do not include in a regular build that you're giving out to others.

That said, feel free to use it for investigating spacetime... just don't completely trust it.  If there's a mistake, it may affect how spacetime beam behaves.


idlechild:
Rolling everything up into one PR.  I also just added a "Freeze on Load State" feature based on discord comments.  It works well enough to use (I think) although it is strange that Samus isn't rendered until you unfreeze.  I'd consider it an experimental feature.

Let me know if there are any objections to stuff in this PR.  In theory it's ready to go.

InsaneFirebat:
I have one more little bug fix I'd like to throw in. Just added it to my pull.
https://github.com/tewtal/sm_practice_hack/pull/84/commits/1cd67ef45d6ff8986dea886929bfb3f4cd126193

It's a fix for decrementing numfields to zero when the minimum value is non-zero.

InsaneFirebat:
It took me a bit to realize that I needed to make the savestate after toggling the freeze feature. The loadstate was turning it back off since it's not stored in SRAM. Was that intentional?

idlechild:
Yes, I intentionally put in as little effort as possible into that feature.  I'm honestly a little surprised you haven't broken it; I guess simpler is better sometimes.

In the future, perhaps we could make it an sram setting, and perhaps at least get Samus to render once while game is frozen, and perhaps make frame advance still work like it normally does if you enter frame advance.  Normally you can enter new inputs as you frame advance, but with this feature toggled on it unfreezes you.

idlechild:
Will probably push all of this in later today, seems like it is ready

PR #87, opened 2022-03-19
Last realtime door

cout:
This change adds a new variable `ram_last_realtime_door` which holds the full door transition time and not just the door "lag" time.  The variable replaces `ram_transition_flag_2`, which was unused.

I am using this in my room and segment timer (https://github.com/cout/sm_room_timer) to accurately measure full room time.  The full transition time is needed so the individual room times can meaningfully be added together to create a realistic estimate of RTA time in a real speedrun.  Replacing `ram_transition_flag_2` instead of adding the variable further down in memory ensures I can easily scrape the variable along with the other room time variables when the transition completes.

This also modifies the full transition lag counter in the infohud to use the new variable (the previous code stored the full transition time in `ram_last_door_lag_frames` when "Transition Lag" was set to "FULL").  This ensures I can still scrape door lag time regardless of what "Transition Lag" is set to.

### DIFF
--- a/src/defines.asm
+++ b/src/defines.asm
@@ -15,7 +15,7 @@
 !ram_last_door_lag_frames = !WRAM_START+$0C
 !ram_transition_counter = !WRAM_START+$0E
 !ram_transition_flag = !WRAM_START+$10
-!ram_transition_flag_2 = !WRAM_START+$12
+!ram_last_realtime_door = !WRAM_START+$12
 
 !ram_seg_rt_frames = !WRAM_START+$14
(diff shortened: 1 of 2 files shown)


idlechild:
FYI after merging in these changes, I'm noticing the time spent collecting an item isn't being counted.  It's possible it's a merge issue on my part or perhaps a different change, but I'm guessing this PR has something to do with it.  I'm getting 2.5x moat CWJ times, which is the same you get if the item is already collected.  I should be getting 4.0x times.

idlechild:
Sorry about the false alarm!

PR #89, opened 2022-04-10
Avoid accidental pauses from controller shortcuts

NobodyNada:
As recently discussed in Discord, controller shortcuts involving the start button (most commonly the Start+Select menu shortcut) will cause the game to pause unless start is the last input pressed. In order to avoid this, check if the game is in the "fading out to pause" mode after handling most shortcuts. If so, cancel the pause by returning to the "normal gameplay" mode and resetting the screen brightness.

This behavior applies to all controller shortcuts except loading presets (because presets reset the game mode anyway) and saving/loading savestates (because the user might actually want to savestate in the middle of a pause, for instance to practice a gravity jump).

I don't *think* there's any way this could break anything, since the only thing it does is abort the fade animation. It should always be safe to set brightness to 15 since the unpause animation always sets it to the max; and I can't think of any scenario where the user should not be able to break out of a pause animation. But this game has surprised me before, so it's definitely possible I missed something.

### DIFF
--- a/src/gamemode.asm
+++ b/src/gamemode.asm
@@ -259,6 +259,8 @@ endif
     LDA $AB : PHA
     LDA #$0004 : STA $AB
 
+    JSR skip_pause
+
     ; Enter MainMenu
     JSL cm_start
 
@@ -266,7 +268,7 @@ endif
     PLA : STA $AB
 
     ; SEC to skip normal gameplay for one frame after handling the menu
-    SEC : JMP skip_pause
+    SEC : RTS
 }
 
 if !FEATURE_SD2SNES
--- a/src/gamemode.asm
+++ b/src/gamemode.asm
@@ -40,6 +40,30 @@ gamemode_start:
     RTL
 }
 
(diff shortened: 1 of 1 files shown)


idlechild:
I'm not sure about this one, seems like we should only ignore the start press if we've been holding start the whole time.  Then again that'd be a bit more complicated, and this is an annoying problem, so perhaps this is a good compromise.

I think the fix for the menu shortcut should be applied before the menu opens since the menu will be open for a while.

NobodyNada:
Hmm, can you clarify what should be the exact conditions for when we should vs shouldn’t skip pausing the game? Would “ignore the pause if start is held at the moment the shortcut is triggered” be sufficient? Maybe we should take into account whether the start button is needed to activate the preset? I don’t have a problem with making the logic as complicated as it needs to be, I’m just not sure what conditions we want.

Moving the fix to before the menu is easy; I’ll implement that once I’m back at my computer tonight.

NobodyNada:
this was merged in #86

PR #90, opened 2022-04-10
Fix music after teleporting/loading a save

NobodyNada:
When we skip the start-game fanfare, we also skip starting the room music at the end of the fanfare. So add the room music back in, in place of the fanfare.


idlechild:
Looks good to me

NobodyNada:
merged in #86

PR #94, opened 2022-05-20
Support for savestates with 128k SRAM

tewtal:
Some devices (like the MiSTer) only supports 128k SRAM and thus doesn't support the main 256k SRAM savestates.

This patch only saves a subset of WRAM/VRAM to SRAM to keep within the 128k limit and instead restores the missing part from the ROM.

Draft PR for now since there's most likely a bunch of issues to work out, and loading back saves needs to be improved to be faster.


tewtal:
Attaching a pre-built IPS of the early test version for easier access to testing, please report back issues in this PR.

[smhack20_tinystates.ips.zip](https://github.com/tewtal/sm_practice_hack/files/8742328/smhack20_tinystates.ips.zip)

idlechild:
%cm_numfield("Custom Preset Slot", !sram_custom_preset_slot, 0, 8, 1, 2, #0)
Should be 7 instead of 8

That was the only issue I saw; looks good.  Happy to see you found another use for preset_load_level_tile_tables_scrolls_plms_and_execute_asm... which actually since you are also storing 7F0000:7F9601, maybe you don't need the "level" part of that method?  You're probably noticing landing site loads a lot slower than a save station room.  Maybe you can cut the level data loading out.

tewtal:
> %cm_numfield("Custom Preset Slot", !sram_custom_preset_slot, 0, 8, 1, 2, #0) Should be 7 instead of 8
Ah okay, I wasn't sure exactly how that worked, thanks for catching that.

> That was the only issue I saw; looks good. Happy to see you found another use for preset_load_level_tile_tables_scrolls_plms_and_execute_asm... which actually since you are also storing 7F0000:7F9601, maybe you don't need the "level" part of that method? You're probably noticing landing site loads a lot slower than a save station room. Maybe you can cut the level data loading out.
Good point, that'd save a bunch of time.

I'm planning to make some checks as well to see what tileset graphics that's already loaded and so on, to reduce the loading of redundant data.

idlechild:
Definitely go after the level data first.  Tileset graphics should be transferred to VRAM fairly quickly; shouldn't be much different from the original loading savestate logic with 256k SRAM which we know is fast.  I'm not sure you'd be gaining much trying to skip that; might be more trouble than it is worth.

InsaneFirebat:
Nice work and great idea. It will be cool to see some of the decompressed data put to more use. This works on every emulator I've got except (banned) Snes9x v1.43. Everdrive is listed as having 128k of SRAM, so I'm guessing it would work there. If this works out on SNES Classic (will be tested tomorrow) and maybe Wii VC, then we wouldn't really need a third build. It could just replace the non-savestate build.

DyceNL:
I can confirm this works perfectly on my MiSTer, awesome work!

tewtal:
I've modified the code now to skip loading the level data, and it also uses the uncompressed data in better ways.
Loading states is now much faster.
Fixed a few other small issues I found as well.
Attaching an updated IPS with the fixes included
[smhack20_tinystates_v2.ips.zip](https://github.com/tewtal/sm_practice_hack/files/8748548/smhack20_tinystates_v2.ips.zip)

InsaneFirebat:
Looks like the BG3 tileset isn't updated. It reveals the hidden debug font for the HUD if you loadstate from game over. It throws garbage text all over the HUD if you load from a crash, which is possible now that we have a crash handler. 

Unfortunately, it messes up rooms with the FX layer, like moat or landing site. Messages boxes are also messed up. Opening and closing the menu fixes it all.

tewtal:
Oh okay, that should be easy to fix thoug hopefully. I'll take a look at it when I can.
I noticed as well there are issues with Kraid corrupting BG2 which doesn't get properly reloaded when using there.

tewtal:
The latest commit should hopefully fix the BG3 issues as well as fixing another issue with mirrored BG2 backgrounds.

Kraid remains an issue where re-loading states while fighting kraid will end up with corrupt graphics. I tried to fix it but haven't figured out a good way to do that yet.

src/tinystates.asm
```diff
@@ -443,3 +444,36 @@ vm:
 
 print pc, " tinysave end"
 warnpc $80FC00 ; infohud.asm
+
+print pc, " tinysave bank82 start"
+org $82FE00
+
+tinystates_preload_bg_data:
```
InsaneFirebat, line 451: print pc should be after the org

InsaneFirebat:
I'm hearing that SNES Classic throws a C7 error with the tinystates rom, so it might be too soon to retire the non-savestate build :(

tewtal:
> I'm hearing that SNES Classic throws a C7 error with the tinystates rom, so it might be too soon to retire the non-savestate build :(

Oh okay, that's unfortunate.

I'll do the changes needed then to support a third option in the website.

src/tinystates.asm
tewtal: Fixed with [28072e2](https://github.com/tewtal/sm_practice_hack/pull/94/commits/28072e250112b7de6813b160eff9a52cf79a6823)

tewtal:
The remaining major issue right now then is loading/saving while in Kraid's room. If someone wants to take a look at fixing that I'd be happy because I couldn't find a good solution. The problem is that since Kraid constantly modifies BG2 it's tricky to find a good way to properly upload the correct graphics.

NobodyNada:
I've taken a stab at fixing Kraid's graphics in #98.

tewtal:
@NobodyNada Thanks for the Kraid fix, all this needs now is some final playtesting and it should be ready to be merged at some point.

Maybe it'd be a good idea to add some documentation explaining the differences more in detail between the three versions now as well.

InsaneFirebat:
I gave it a shot with PR #102

I would have liked to add a line above the Platform drop-down box that said something like, "The platform you play on determines how the built-in savestate feature works, or if it can be included at all." but the javascript was beyond my understanding. I don't even know how to test such a change before committing.

InsaneFirebat:
There's still a bug with BG2 scrolling on certain backgrounds. Load up a `Maridia > Fish Tank` preset, save a tiny state, and then load it to see the issue. I've checked all of the BG2 scrolling and tilemap addresses I could find, and I don't see anything out of place.

InsaneFirebat:
Added another commit to pull #103 that adds $200 bytes of `!WRAM_START` to the save/load DMAs. That addresses the issue with the segment timer and some other inconsistencies compared to normal savestates.

NobodyNada:
I've spent some time investigating the BG2 issue. The root of the problem is simply that tinystates do not save the contents of the BG2 layer.

The `tinystates_mirror_bg_data` routine is a workaround that allows us to reload BG2 data, but it only works in small rooms (smaller than 20 square screens). The routine is copied from a part of [the routine the game uses to load level data](https://patrickjohnston.org/bank/82#fEA73). The original routine works like this: first, the game decompresses the BG1, BTS, and BG2 layers into bank $7f; then it relocates the BTS and BG2 layers to start at the correct addresses (since the decompressor simply places the one after another).

When loading a tinystate, we run only the "relocate the BG2 layer" part of that routine. This is problematic, because it assumes a copy of the BG2 layer is still present in the temporary location where the decompressor originally put it -- it's essentially a use-after-free bug. It happens to work for small rooms, because as long as the room is smaller than 20 screens the entire level fits within the space allocated to the BG1 layer, so `tinystates_mirror_bg_data` manages to recover the BG2 layer from the unused memory allocated to BG1. But for larger rooms this fails, since the memory used to temporarily store BG2 after decompression overlaps the buffer used to store BTS data post-decompression.

I'm not sure what's the best way to fix this. As I see it, we have two options:
- Save the BG2 layer in SRAM (in addition to BG1 and BTS).
  - This would mean we'd have to cut down on about $6000 bytes of SRAM somewhere else.
    - Maybe we can save less VRAM data? For instance, it seems like we're saving $4000 bytes of BG3 VRAM, which (as far as I can tell) is entirely blank except for the status bar on top.
- Reload the BG2 layer from the level data when we load a savestate. This would mean decompressing the entire room, which takes a second or two. But maybe we can add some optimizations:
  - If the room is small enough to safely load BG2 from unused memory (like we currently do), we don't have to decompress the room.
  - If we're reloading the room we're already in, we probably don't need to reload anything.

InsaneFirebat:
Nice work. What are the consequences of skipping `tinystates_mirror_bg_data`? Removing that call fixes the BG2 issues and I'm not seeing any of the bugs mentioned before it was added.

NobodyNada:
My understanding is that BG2 would not be updated at all upon loading a state. I don’t have time to test it right now, but here’s an experiment you can try:

- Go to Landing Site and make a savestate.
- Go to Green Brinstar Main Shaft.
- Load the savestate.
- Run around in landing site so off-screen areas of BG2 are loaded into VRAM. If I’m understanding things correctly, you’ll see the wrong background tiles since the game is now using the BG2 layer from GBMS.

InsaneFirebat:
It seems the scrolling sky background is immune to that issue, as are many smaller roomers. I did eventually see the bad backgrounds though.

As a workaround, we could check for the four problematic rooms and use the slower method of loading from the compressed data. Would look something like this: https://github.com/InsaneFirebat/sm_practice_hack/commit/d2ecc1d8696a9c9661d3101eafd8ffb0a3688807

NobodyNada:
After some experimentation, I think the best course of action is to always load from the compressed data, unless we're loading a savestate from within the same room (in which case we don't need to reload the level at all). My patch is here: https://github.com/NobodyNada/sm_practice_hack/commit/fc0ed923d1515a5ff7927dcac2cee7a5ca8d3806

I tried logic similar to yours that would skip decompression for small rooms, but I found this to not be worthwhile because it only takes a couple frames to decompress a small room. The added code complexity of deciding whether or not we need to decompress the room is IMO not worth the imperceptible performance gain.

InsaneFirebat:
I've come across a few bugs in recent testing. These occur when loading from one boss fight into another:
- Phantoon's BG2 tilemap gets scrambled
- Draygon's BG2 tilemap is either facing the wrong direction or scrambled
- Ridley's tail is doing weird shit, or the savestate crashes until the game is reset

All of these are fixed by storing another $1000 bytes from $7E2000. I don't know how much of it we should actually be saving, but all of it is used for corpse-rotting graphics in baby room, which can also be a bit scrambled. I've made this fix for my custom builds, where I also moved CGRAM and the savestate defines for tighter organization. That leaves $1000 bytes in bank $73 and at least $800 (I use $200 more in my custom branches) in bank $70.
https://github.com/InsaneFirebat/sm_practice_hack/commit/18b7031d6af8dc627e2aeca368bc9e0bdd9eda6f

Spreadsheet for SRAM usage: https://docs.google.com/spreadsheets/d/1E1cAbN94-MMX6rx3pV-TyEDMiVmV500GsRzyENFdWLo/htmlview

PR #97, opened 2022-05-25
Improve Extended Spritemap Hitbox Viewer

InsaneFirebat:
I've made some changes to the extended spritemap hitbox viewer (mostly for bosses) to properly handle spritemaps from Mother Brain, Ridley, and Crocomire (maybe others). It now checks for multiple spritemaps, and also applies X/Y offsets from the spritemap itself. Most other bosses had zero offsets in their spritemaps, so they were still drawing correctly before.

Mother Brain, Ridley, and Crocomire have additional custom hitboxes. I added an option to draw these for Mother Brain and Ridley. Crocomire's extra collision is not a traditional hitbox, but rather a routine that does some quick X position math (this routine is skipped while Croc is hurting from chain-plasma damage).

The HUD image on the help page (https://smpractice.speedga.me/help) has been updated and now includes the Vanilla HUD layout.


InsaneFirebat:
Added some comments to menu.asm and reorganized some things before and after the menu loop. I also shifted the number fields, toggles, and choice menu items over a bit to maximize our space for naming things. There's still one tile of padding before the border on each side.

`build_dev.bat` will now output debugging symbols in WLA format. You can import them into Mesen-S to get labels for just about everything inside an org (not WRAM).

InsaneFirebat:
I discovered that the hijack for updating timers at boss drops was running inside of a loop, and wasting around a third of that frame updating the HUD which it should only take a few scanlines. I also noticed that we're pushing/pulling the working registers unnecessarily in `ih_update_hud_early`.

InsaneFirebat:
The HUD init routine was resetting IGT between Ceres and Zebes, which makes it impossible to get an accurate game time with a full playthrough. Anything that needs to reset IGT (presets/teleporting) handles that already.

PR #99, opened 2022-07-06
Add an option to remove liquid from the current room

NobodyNada:
For practicing gravity jumps in randomizers:
https://discord.com/channels/98929157894836224/98929157894836224/993351592684310599

### DIFF
--- a/src/infohud.asm
+++ b/src/infohud.asm
@@ -1321,11 +1321,20 @@ magic_pants:
 +   RTS
 
   .check
-    LDA $0A1C : CMP #$0009 : BEQ .flash
-    CMP #$000A : BEQ .flash
-    RTS
+    ; Enable magic pants in the following states:
+    ; 9: Moving right - not aiming
+    ; Ah: Moving left  - not aiming
+    ; Bh: Moving right - gun extended
+    ; Ch: Moving left  - gun extended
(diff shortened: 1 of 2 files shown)


InsaneFirebat:
There is a `cm_toggle_bit_inverted` macro you can use here, assuming you want to allow turning the water back on. The `_bit_` part makes it an EOR and the `_inverted` part is for drawing ON/OFF the opposite way.

InsaneFirebat:
I went with "Disable Water Physics" but "Disable Water this Room" should just barely fit. If it doesn't, I've been meaning to adjust some of the drawing positions in our favor anyway. Maybe soon, but could be now if the need arises.

I would avoid using "liquid" since it only affects water physics (also affects space jump leniency). Lava and acid appear to behave the same.

NobodyNada:
Hmm, "Disable Water this Room" is just one character too long. "Ignore Water this Room" fits (though admittedly it's a little cramped).

NobodyNada:
I pushed a change to enable magic pants while in running-while-shooting-or-aiming-diagonally poses, as requested on Discord: https://discord.com/channels/98929157894836224/890883815449444372/994413550309216266

Since this is such a minor bugfix I stuck it on here instead of making a new PR for it.

PR #100, opened 2022-07-06
 Add option for fast doors & elevators 

NobodyNada:
Doors are implemented by simply running the regular door transition code twice; that way we don't need to deal with things like correcting Samus's position because we're running the same code the same amount of times as usual. 

This patch only affects the speed at which the screen scrolls from one room to the next; it does not change the fadeout/ddoorfix/processing time at all. This means the door time shown in the HUD will always be the same (except on a downwards elevator, where enabling fast doors reduces the door time by 32 frames because the elevator delay is reduced.)

Marked as a draft because I have not yet tested this thoroughly, and because I still need to update some of the addresses for PAL. The only issue I know about is that some vertical doors flicker slightly because decompressing the next room takes too long (this is purely a graphical issue). VARIA has [a "fast decompression" patch](https://github.com/theonlydude/RandomMetroidSolver/blob/master/patches/common/src/decompression.asm) to fix this; though I did not include it because:

1) [It seems to have some bugs](https://discord.com/channels/533418016218677260/676518457373622322/965708043067228243)
2) VARIA is licensed under MIT, which is not compatible with the practice hack's public domain unlicense. This just means we'd have to obtain permission from Kejardon to distribute the patch

This PR also speeds up the Nintendo splash screen on boot.


InsaneFirebat:
Can the fast logo be optional in the cutscenes menu? I love the idea but I'd rather have it as normal myself.

NobodyNada:
I’ve found a bug: for whatever reason, when entering Mother Brain’s room with fast doors enabled, the two rinkas spawn simultaneously instead of one after another, making downback zeb skip impossible.

NobodyNada:
I've fixed this bug. Turns out Rinkas are sensitive to how far the screen is scrolled when they're initialized, and that depends on how far into the door transition the tileset decompression finishes. Fast doors meant the door scrolled further before decompression finished, which in turn caused the Rinkas to spawn in a different order.

I fixed this by turning on fast decompression when fast doors are in use. I still need to fix this up for PAL and write documentation, but once that's complete this should just about be ready to merge -- I'm not aware of anything else that's broken.

PR #101, opened 2022-07-07
Removed the Blue Suit option for preset_hundo_draygon_botwoon and set_hundo_draygon_full_halfie.

lightquarz:
Removed the blue suit part of 100% late croc.
Please confirm.


lightquarz:
thanks
I've also fixed the AllBosses category.
I feel a little weird that the settings are applied to Terminate.

InsaneFirebat:
> I feel a little weird that the settings are applied to Terminate.

The terminator presets are special because of the camera movment when approaching the door. I wanted players to be able to setup their Terminator entry the way they would in a run, so I made those presets start on the other side of the bomb blocks. To get through the bomb blocks quickly, I added a blue suit. It seemed like the easiest method to address the issue with the knowledge I had at the time.

lightquarz:
There was such a background.

InsaneFirebat (approved):

PR #103, opened 2022-07-17
Fix some crashes from bad savestates

InsaneFirebat:
This pull targets the tiny-state branch (PR #94). Tiny states will crash on gamemodes 0-6 and most of $1C and beyond. Some of the gamemodes between $07-1B have garbage graphics (pause menu) but are allowed because they don't crash. I allowed $20 since it loads fine and gives the player a chance to save the escape timer after touching the elevator, up until the fade out.

The other fix is for both types of savestates. The "5AFE" word is written to SRAM at the end of `save_return`, and will be checked before loading a savestate. Both checks are performed before jumping to save/load routines so that they don't skip a frame of gameplay when returning to `gamemode_shortcuts`.


InsaneFirebat:
Merged with the 2.5 rollup.

PR #109, opened 2022-09-20
Labels for menu DP usage

InsaneFirebat:
This pull is mostly about labeling our DP variables for better code readability. There are also some new labels for a few vanilla addresses, including the HUD tilemap. I've made this exact change three times now in other branches/projects, but there's still a decent chance I missed something. Extensive menu testing is needed.

I've also remade the equipment/beam menus so items can be unobtained, unequipped, or equipped. It required some routines and new variables to setup those menus, but it reuses the same chunk of RAM that the RAM Watch menu would use. There's also a couple of new menu macros to make all those ACTION_CHOICE menus easier to read/write.


InsaneFirebat:
I'm gonna push this to the next-update branch soon if no one has any issues with it.

PR #116, opened 2022-10-04
Experimental infinite-doppler mode for Phantoon.

NobodyNada:
When Samus presses L+R during a Phantoon swoop, record her positon. Each time she fires a missile, teleport her back to the initial position. Each time a missile hits, teleport it back by the amount Samus moved before firing it, and allow it to hit again. This way, Phantoon's cooldowns behave as if we were dopplering,

A proof-of-concept. Doesn't have a UI yet, or a way to turn it off.

### DIFF
--- a/src/defines.asm
+++ b/src/defines.asm
@@ -90,7 +90,20 @@
 !ram_quickdrop_counter              = !WRAM_START+$7A
 !ram_walljump_counter               = !WRAM_START+$7C
 
-!WRAM_PERSIST_START = !ram_walljump_counter+$02
+!ram_infidoppler_enabled         = !WRAM_START+$7E
+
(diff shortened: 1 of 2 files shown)


InsaneFirebat:
Gonna leave this pointed at master until it gets more activity.

NobodyNada:
Finally got around to finishing this.

If the menu option is turned on, infi-doppler activates automatically when you fire the last missile during a swoop, and deactivates once the swoop ends (so there's no more weird shortcut to trigger it).

This used up the very last byte of free WRAM -- we'll have to refactor/rearrange the next time we add a feature that takes WRAM.

https://github.com/user-attachments/assets/0fcab987-092e-4a6a-ae41-90a86f790b57

PR #119, opened 2022-10-12
Ridley AI room strat

InsaneFirebat:
This room strat identifies Ridley's AI pointer and prints text on the bottom-right portion of the HUD to name the routine. After the text, Ridley's HP is drawn since that portion of the HUD isn't being used. When Ridley's HP is zero, it will use that space to count Ridley's death grab attempts instead.


InsaneFirebat:
@Yuriks proposed this solution https://gist.github.com/yuriks/e8c5d2605501bed488d9c4d613da5e54 which is a good bit more optimized for speed. I had to make some adjustments since our data bank is set to $80, and it didn't account for pre-fight pointers.

It uses the high byte of the pointer to find a starting offset into the table, so the pointer should be identified in 4 checks or less. The index is set to the max value for unnamed routines so only one check is wasted before grabbing the fallback text pointer,

PR #129, opened 2022-10-31
Unnecessary update

idlechild:
Pulled in misc-2.5.1-things branch
Now creating two debug symbols files (one sorted by name and one sorted by address)
Added labels next to all of our patches so they show up in the debug symbols
Named all of the debug symbols (replacing +/++/- labels)


idlechild:
lol
Okay I renamed my PR, please don't accept it as is.  I made a separate PR with the small updates I had made plus your misc-2.5.1-things branch; you can take that one.

I would port in all the new hook labels so that the debug symbols includes all the places we have patched vanilla code, but I need to find a better diff tool so I can sort out those changes from the +/- changes.  Also it's just not that important; I may not get back to it for a bit.

I'm hopeful there's a compromise regarding the labels.  What's the longest a label can be before it becomes annoying to read?

PR #147, opened 2023-02-12
DASH Recall Layout Patches

idlechild:
Work in progress.  Can share if people want it, but not ready to officially release it to 2.5.3 yet.

Pants Room has a visual glitch that changes depending on which door you enter from, not sure what's going on there.


InsaneFirebat:
That last one is fixed in the `next-update` branch

idlechild:
This is now *very* close to completion.  I'm hoping to update the snail clip room strat (it's more difficult without hi-jump boots) and then update the help pages and then it is done.  Ideally I'll push it tomorrow morning.

Also I'll be on vacation starting tomorrow afternoon so if there are problems I'll need help.  Everything has at least been briefly tested (I've loaded every room that I modified).  Some of the changes have been beta tested by others as well.

idlechild:
This PR is done!  I'll leave it to you (InsaneFirebat and/or NobodyNada?) to decide when to pull it in.  It's ready now, but of course you're welcome to do some testing first or include more in the 2.5.3 update.  I'll be on vacation for the next six days including today; I'll be on discord at times but not able to test or fix anything.

NobodyNada:
Nice! I assume we'll want to revert #99 since the new water physics option can be set to "off to on" to do the same thing?

Also, because of the toggling option do you think it would make more sense to have water physics in the non-persistent section of WRAM, so that it's restored to wherever you had it when you load a save?

InsaneFirebat:
I added a fix for the new !SAMUS_WATER_PHYSICS variable being cleared when teleporting to a save station. Also removed the old water physics option that NobodyNada mentioned.

I'm not sure about the persistence suggestion. I'd think a rando player would want it to stay on the rando setting when they boot up, but I'm just guessing.

Issue #108, opened 2022-09-13
Door Bonk Tool


InsaneFirebat:
Suggestion from ShinyZeni:

> Probably mostly useless but for us Ceres grinders I’ve been wondering about the viability of a door bonk tool. Like how many frames were left before the door would have disappeared when Samus bonked.
> Even better if it could be combined with the Arm Pump tool lol

Arm Pump tool (by idlechild):
https://github.com/InsaneFirebat/sm_practice_hack/blob/CustomROMBase/src/infohudmodes.asm#L1121

InsaneFirebat:
Not sure which pull it went into master with but it's completed as mentioned above

PR #150, opened 2023-03-08
Door bonk tool

NobodyNada:
When you  bonk a door, displays the number of frames before the door would have finished opening. This is displayed to the left of the lag counter (and only if the minimap is disabled).

Screenshot 2023-03-07 at 3 58 46 PM

When the game zeroes Samus's X speed after a horizontal collision, we add a hijack that does the following:

1. Check to see if the collided tile was an opening door
2. Loop over all PLMs, until we find a door PLM that has drawn to the collided tile
3. Loop over that PLM's remaining draw instructions before it makes the door non-solid, adding up the animation delays.

Does not take into account the fact that the middle of the door opens before the bottom.

TODO:

- [ ] PAL support
- [ ] Ceres support (probably just needs a couple changes to constants, but I haven't looked into this fully)
- [ ] Support for vertical doors (already supported by the routine, so it should just need some hijacks in the right spots)

### DIFF
--- a/src/infohud.asm
+++ b/src/infohud.asm
@@ -73,6 +73,13 @@ endif
     JSL ih_babyskip_segment
 
 if !FEATURE_PAL
+    ; TODO
+else
+org $90E5E9
+endif
+    JSL ih_door_bonk_tool
+
+if !FEATURE_PAL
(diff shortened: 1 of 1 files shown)


NobodyNada:
Resolves #108

PR #156, opened 2023-05-11
Practice Menu Customization

InsaneFirebat:
Custom palettes and sound effects for all!

This pull adds a new submenu with options to disable the menu background, edit the palettes, and customize the sound effects in the practice menu. This is strictly a "for fun" update and adds no gameplay/practice advantages.

I also added total's RNG routine from Arcade. Its seeds are stored in SRAM until the menu is opened, and it'll cycle every frame as part of cm_loop. The seeds are stored back to SRAM when the menu is closed. I added this for the palette randomizer option which needs to get a lot of new RNG numbers, and I didn't want it cycling vanilla RNG in the process. It's also used for loading random presets.

Since I've had this in my custom builds for a while, I've accumulated a fair number of palette profiles from community members. There's a non-interactive menu in there to "Share Your Colors" which instructs the user to take a screenshot and submit. We may get requests to add more color profiles in the future. They can be moved around, so size isn't an issue, but we could potentially get too many requests. Open to feedback on all that, or whatever you've got.


InsaneFirebat:
I got my folders mixed up and was test building the wrong repo. Ready to merge this now.

PR #159, opened 2023-05-14
Added menu to inspect custom presets

InsaneFirebat:
This pull turns the numfield for custom preset slots into a "hidden" submenu that displays information about saved preset slots.

Screenshot of new preset slot menu.


src/mainmenu.asm
```diff
@@ -412,10 +412,28 @@ presets_goto_select_preset_category:
...
 else
-    %cm_numfield("Custom Preset Slot", !sram_custom_preset_slot, 0, 39, 1, 2, #0) ; update total slots in gamemode.asm
+    %cm_numfield("Custom Preset Slot", !sram_custom_preset_slot, 0, 39, 1, 2, #.routine) ; update total slots in gamemode.asm
 endif
+  .routine
+    ; ignore if not A, X, or Y
+    LDA !IH_CONTROLLER_PRI_NEW : BIT #$40C0 : BNE .submenu
+    RTL
+  .submenu
+    LDA !sram_custom_preset_slot : BEQ .zero
```
InsaneFirebat, line 424: Should have put a comment here. Pressing A on the numfield will increment the slot before landing here, so we undo that before jumping to the hidden submenu.

InsaneFirebat:
Opening the hidden menu causes the move and number sound effects to play consecutively. Not sure if I'll fix that or not.

I'm happy with this pull now.

InsaneFirebat:
Bank $E7 now contains room names which can be found using the room ID as a pointer. Users can press Y in the preset slots menu to switch between viewing Samus stats and room names.

InsaneFirebat:
I wanted to have the cursor jump to the current preset slot when the submenu was opened, but it creates a bug where the previous cursor index can be out of bounds upon returning. This is because the extra pages are treated as the same menu depth to avoid players flipping through pages and overflowing the menu stacks. I'll settle for loading the correct page instead.

I'd say this is ready to go again.

PR #160, opened 2023-05-14
Add option to show projectile sprites as 32x32

idlechild:
Todo:  Rework sprite features RAM to use bits so we can free up some WRAM_PERSIST addresses.  May not get to that for a little bit.

Feel free to share or incorporate with other work.  I only made this a draft PR because of all the other work going on; not expecting the 2.5.5 update to happen just yet.


InsaneFirebat:
> Todo: Rework sprite features RAM to use bits

Mind if I take care of that? I've already done it in a separate commit for my custom builds.

idlechild:
32x32 mode makes sense during botwoon fight now.  Also added an optimization for 32x32 mode (don't draw next projectile if previous projectile was in same location).  Also added an optimization for regular mode, if both X and Y radius are below min_four_corners_radius then only draw two corners.

The "correct" value for min_four_corners_radius is 5, but I found setting it to 7 reduces the number of sprites during MB rainbow beam, so I set it to 7... but I'm not sure which is better.  Feel free to change it back to 5.

idlechild:
Oh also let me know if this PR should be re-routed to next update branch, or just go ahead and pull it into next update branch.  I don't have any more updates planned.

PR #161, opened 2023-05-21
v2.5.5 Update

InsaneFirebat:
Pulling together the changes for the next update. Still needs the 32x32 projectile sprites pull before it goes live.


src/init.asm
```diff
@@ -88,7 +93,8 @@ init_sram:
     CMP #$000E : BEQ .sram_upgrade_EtoF
     CMP #$000F : BEQ .sram_upgrade_Fto10
     CMP #$0010 : BEQ .sram_upgrade_10to11
-    JSR init_sram_upto9
+    CMP #$0011 : BEQ .sram_upgrade_10to11
```
idlechild, line 96: Copy-paste strikes again

idlechild:
> Pulling together the changes for the next update. Still needs the 32x32 projectile sprites pull before it goes live.

@InsaneFirebat Did you have an optimization for the sprite WRAM, using bits so they all share the same word of memory?  Just wondering if you want to pull in that 32x32 update and then add that optimization.  Or you could bring in that optimization and I'll bring in 32x32 on top of it.  Or you could do neither and I could just do that optimization since I don't think it would be much work.

Regardless, I might have more changes I want to make for that, perhaps make the sprites bigger so they are actual boxes and maybe change the color.  I also suspect there's a simple optimization I could do for botwoon use case where if we've already drawn a box and the next box is in the same 32x32 square then skip the box.  No need to iterate the whole list so it wouldn't be perfect but still an improvement.

Just letting you know once 32x32 gets merged I might have a bit more I want to do before it ships.

src/init.asm
InsaneFirebat: Thanks, that's fixed now.

src/custompresets.asm
InsaneFirebat, line 1: This can be reverted since area is no longer part of the custom preset display.
Remove !PRESET_SLOTS_AREA define as well.

src/macros.asm
```diff
@@ -44,3 +79,283 @@ endmacro
...
+endmacro
+
+macro cm_custompreset(slot)
+; Displays custom preset data and allows selecting active slot
+custompreset_<slot>:
+  .dm_actionIndex
+    dw !ACTION_CUSTOM_PRESET
+  .dm_slot
+    db <slot>
+  .dm_text
```
InsaneFirebat, line 251: Debugging labels can be removed

InsaneFirebat:
I'm thinking about adding a new Cutscene option to skip the game over screen. If enabled, it would try to load a savestate, then a preset, then fallback to game over. The Lion King community requested it for their practice hack, but I wasn't using built-in savestates at the time and didn't realize it would be nice in SM too. It could avoid the music change delay.

InsaneFirebat:
> The "correct" value for min_four_corners_radius is 5, but I found setting it to 7 reduces the number of sprites during MB rainbow beam, so I set it to 7... but I'm not sure which is better. Feel free to change it back to 5.
I agree. It looks nicer at 5 but it's better to have ketchup beam fully visible. I'd leave it at 7.

Docs are included now. Everything seems good to me.

Issue #148, opened 2023-02-26
100% Early Croc Leaving Phantoon preset shows gray screen


EXAKT-Science:
"Leaving Phantoon" preset in 100% Early Croc doesn't display the room properly, with Samus and all tiles but the door being covered with horizontal gray bars. Bomb and PBs will display, but no other projectiles. The "Samus on Top" Sprite Features setting will show her properly, and using X-Ray will show the room normally, even outside of the X-Ray cone. Screenshot attached.
![leaving_phantoon_preset_bug](https://user-images.githubusercontent.com/119459340/221433105-f72d793c-99a1-4bc3-b928-bc8479f41d52.png)

InsaneFirebat:
This also affects any preset that starts in Bowling Room. Custom presets as well.

InsaneFirebat:
Apparently this doesn't affect Bowling Room anymore. [b5ee6a7](https://github.com/InsaneFirebat/sm_practice_hack/commit/b5ee6a7ede6196532de7eee63a81cc7402fa22d0) is a hacky fix to clear and reload BG2 if a preset loads into Phantoon's room.

PR #162, opened 2023-05-22
Auto-Savestate in Doors + Minor Commits

InsaneFirebat:
There is a single frame of door transition where execution returns to the main loop and a controller shortcut can be (frame perfectly) used. I've setup a flag that can automatically trigger a savestate on that frame, which is usually right in the middle of door scrolling. There is a controller shortcut, as well as menu option, to set the flag. Since we have a few savestate related settings now, I made a Savestate Settings menu for Rerandomize, Freeze on Loadstate, and Auto-Savestate Mid-Door.

We had a question in Discord about what the 2 mean in some of the Ridley AI function names, so I added a 2 in the same style font. Hopefully that makes it more clear that it's part of the AI name and not a value.

I turned on power bombs for the any% attic presets, since they all start in position for the power bomb strat.


InsaneFirebat:
Realtime Lag Counter is a HUD mode that shows lag frames counting up in real time, instead of waiting for the door transition. It can be useful for identifying which actions are contributing to lag in bigger rooms. It overwrites an unneeded instruction in the lag branch of NMI to increment a counter, and that counter is reset in `ih_after_room_transition`. It adds a few cycles per lag frame, but is otherwise CPU neutral during normal gameplay.

A new Cutscene option allows the game over screen to be skipped. It will first try to load a savestate. If it doesn't exist, it will check the last preset. If that's set to zero, it'll load from the save file. And since it's possible to play without setting up any save files (teleport after boot), we have the final fallback of loading the game over screen.

InsaneFirebat:
Issue #148 strikes again in Draygon's room. Fixed with the same routine to clear BG2 tilemap.

I finally remembered to put in a fix for the crash viewer automatically jumping to a different page when it first runs. Just needed to grab the newest inputs and save them as the previous inputs during initialization.

idlechild (approved):

PR #163, opened 2023-05-28
Add docker support

idlechild:
You can now use docker to build the IPS files.  Also now I can build useful Debugging Symbols.

This could be pushed now since it doesn't change anything.


idlechild:
Ha ha, turns out all I really needed to do was upgrade my local copy of asar from 1.71 to 1.81.  Now I'm getting better debug symbol output without the docker stuff.

It's still not bad to have it though.

InsaneFirebat:
Thanks for making the symbols script. This is going to be a big improvement in the debugger, without the headache I was going through updating it manually.

I just committed wram_symbols.asm to the pull. symbols.asm still needs to be incsrc'd somewhere. Could go in main.asm or defines.asm, but it needs to be after our defines or asar will get mad.

idlechild:
Oh didn't even realize these files were meant to be included in our build, thought you were just loading them into a debugger.  Looks like including at the end of main.asm works.

Also made some improvements to the sorted symbols output.

InsaneFirebat:
> Oh didn't even realize these files were meant to be included in our build, thought you were just loading them into a debugger.

asar will pick them up and add them to Debugging_Symbols.sym for us. Previously, all of our RAM addresses were unlabeled in the debugger.

InsaneFirebat:
asar devs taught me a trick. You can add `> filename.log` to the launch parameters to have it move all the `print pc`'s out of cmd prompt and into a file. Would look something like this, if you're interested.

`..\tools\asar > ..\build\print_pc.log --no-title-check`

idlechild:
I like seeing the print pc output though, at least sometimes it's useful

Ofc if you only change the linux command and leave windows alone I won't care much

InsaneFirebat:
I talked with them some more after originally posting that and what's listed up there now prints everything in both places.

idlechild:
Combined symbols file over 25K lines, and I don't even think it's complete.  That's a lot of edits and symbols!

The goal is to have every single org be reflected in the symbols.  As it is now, if you do an org followed by db/dw/dl with no label then it won't show up there.  I think we have some of those so just need to track them down.

Btw you're welcome to change the build output.  Now with this combined symbols file I will care less about the start/end output.  Having the errors show up at the end sounds nice.

src/crash.asm
```diff
@@ -44,22 +44,31 @@ org $808573
...
+EmulationCOPHandlerHook:
+    dw COPHandler
+
 ; Hijack emulation IRQ/BRK vector
-; Don't do this. A bug in vanilla code causes this to make Mini-Kraid cry
+; Due to a bug in vanilla code, MSB needs to be unchanged (0x85)
 ; https://patrickjohnston.org/bank/A6?just=9BB2
-;org $00FFFE
-;    dw BRKHandler
+org $80FFFE
```
InsaneFirebat, line 64: The bug still occurs with this change since bank $00 is a mirror of $80.

idlechild: It does?  I was thinking the AND $00FFFF would pick up the $85 out of $8577, same as it would if we left it at the vanilla $8573.

InsaneFirebat: Oh, I see what you did below it now. Unfortunately, I can't test it again atm.

PR #165, opened 2023-06-04
Adds option to not update timers when riding elevators or grabbing items

LakeIshikawa:
I needed something like this for better compatibility with sm_room_timer.
The room timer script fetches the room time from memory, but it might happen that it does so with a bit of a delay (>0.25s).
For most rooms this isn't an issue, but for rooms that have elevators or items right after the door transition (like after pit room or the hi jump entrance) this sometimes results in the wrong time being fetched by the script.
With this new option, one can now turn on/off the timer update when grabbing items and riding elevators, which minimizes the room timer issue.
(Naming of the option can likely be improved...)

### DIFF
--- a/src/defines.asm
+++ b/src/defines.asm
@@ -207,6 +207,8 @@
 !ram_seed_X = !WRAM_MENU_START+$60
 !ram_seed_Y = !WRAM_MENU_START+$62
 
+!ram_timers_autoupdate = !WRAM_MENU_START+$64
+
 ; ^ FREE SPACE ^ up to +$7E
 
 ; ------------------
--- a/src/infohud.asm
+++ b/src/infohud.asm
@@ -178,7 +178,8 @@ print pc, " infohud start"
 ih_get_item_code:
 {
     PHA
-
+    LDA !ram_timers_autoupdate : BNE +
+    
     ; calculate lag frames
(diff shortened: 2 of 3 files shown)


InsaneFirebat:
If you use the %cm_toggle_inverted macro on the menu item, it'll default to ON (when zero) and then it won't need to be reinitialized at boot. It would also let you `LDA !ram_timers_autoupdate : BEQ .done` to speed up the check.

ih_get_item_code needs to preserve A for the overwritten routine. Our code needs to fit inside the PHA and PLA in that routine.

LakeIshikawa:
Thanks for the tips!
I updated the PR with all the suggestions (actually had to use BNE for the check but the trick works either way).

As for which events exactly it should be applied to... this is probably something to be discussed with other users as well, but I'll give you my perspective: this new option functions as a workaround to the room timer script fetching the wrong time in cases where a time update happens immediately after a room transition.
This mostly happen in elevator rooms, or rooms that have an item immediately after the door.  I think it will never be an issue for boss drops and Mother Brain timer updates, as those happen long after the transition into the room has occurred.
Not sure if perhaps the behavior should be aligned anyway for consistency?

Unfortunately I predict this won't fully solve the issue if there are rooms which are short enough (around 0.25s) from entrance to exit door, which I haven't come across yet but I guess in 100% or other categories might happen...

PR #167, opened 2023-06-27
Label all patches and reduce label lengths

idlechild:
Bringing back the plan to remove + and - symbols, but this time going after label length as well.  Excluding preset menu/data labels, the longest label is now 52 characters, which is probably still too long.  @InsaneFirebat what's a good target label length for your mesen debugging?

Another advantage this time around is with the asar 1.81 addr-to-line labelling, a lot of our one-line patches automatically show up without needing to add a label.  The first time I proposed this I added labels for all of those patches and that did not help the long label situation!  We still need labels before db/dw/dl commands if those are by themselves.

For the larger commits, I tested dev patches before and after to verify there were no actual changes... but in the shorter commits I did make some changes.  I haven't tested the last change yet for the decompression.  I noticed we always use fast decompression with fast doors?  That makes sense, but does Draygon's room end up messed up?  Perhaps after Draygon is dead and re-entering?  I've heard something turns red.  Anyway with both PJBoy's and Maddo's fixes, hopefully Draygon room is fixed while also not breaking the BTS in other rooms (iirc when loading west ocean you would just fall through the floor).


InsaneFirebat:
> what's a good target label length for your mesen debugging?
Honestly, just do whatever. I'm not a fan of this endeaver but I don't care to fight it anymore.

idlechild:
Merged in several PRs.  For the keyboard PR the sram custom header offset looked wrong so I moved it, hopefully that's fine.

Merging this into next-update branch.  Perhaps ready for 2.5.6 update?  I don't have anything else to add.  I'm not sure how helpful or realistic the fast bowling option is, but I included it.

PR #169, opened 2023-07-10
Custom Door Portals

idlechild:
Open for beta testing, but not quite ready.  Still some issues.  Some obvious (Kraid room no surprise there).  Some not so obvious... something's not quite right with it.


idlechild:
I think custom doors is done.  I'd like to roll out release 2.5.6 soon.  Anyone have anything else that should get in?

PR #182, opened 2023-09-03
Debug Settings menu + misc bug fixes

InsaneFirebat:
Game menu was looking a bit full, so I made a new submenu for it to house the debug related options. Also added the non-duplicate debug load stations to the Teleport menu and separated them by area.

I got a request for resetting the segment timer on item touch, so I added that to the InfoHUD menu and made the controller shortcut work for item touch or exiting the next door, whichever comes first. The timer options in InfoHUD could probably have their own submenu in there too. I might do that later if there are no objections. Lemme know.

There are several small bug fixes included:
- The thousands digit of enemy HP (and other HUD modes) would be overwritten by the vanilla health routine when taking damage using the Vanilla HUD option
- Door alignment and Lag Counter were writing to the wrong place in the HUD when the minimap was enabled
- Mt Everest presets tend to be standing on the bottom left door, which causes Samus to fall through immediately when the door is opened by preset code. Setting Samus to a facing forward pose keeps the door closed.
- My last fix for shinespark presets only addressed a bad branch instruction. Completely missed the fact that it was checking for the wrong poses too. Should be working with all presets now.


idlechild:
Looks good to me.  I just changed your PR to point to next-update instead of master.

InsaneFirebat:
This is ready to go, but I can't fix the merge conflicts on the symbols files. I totally regret suggesting these be added to the repository xD

idlechild:
We should just delete those files and then add them to .gitignore so they aren't in the repo.  I really like having the files but they're easy enough to build locally and not worth the merge conflicts.

NobodyNada:
Maybe a better approach is to configure GitHub Actions to save them as build artifacts. I've already set up artifacts for getting easy "beta" builds from pull requests, would be a tiny change to the YAML to save the debugging symbols as well.

InsaneFirebat:
I'm good with all of this. Can you make a PR for us to sort that out? I'm going to remake this pull so this mess goes away in the meantime.

InsaneFirebat:
Closing. Replaced by #184

PR #183, opened 2023-09-03
First pass at supporting horizontal mirroring of area/boss portals

idlechild:
I need to find a better place to do the mirroring.  It should not be done until after enemy movement is processed (after the fadeout).  I just couldn't resist pushing up a PR in its current state.


idlechild:
Pushing into next-update branch in case others want to test it.

On my todo list is investigating two cries without springball and Zeni's frame-held mode idea.  After that might be ready for another release, including IFB's stuff, anything else we should add?

I'm not going to bother investigating the xray frame advance issue lol.

PR #188, opened 2023-09-09
Don't show "development build" footer for release builds

NobodyNada:


InsaneFirebat:
idk how to test this other than to push it so we're pushin' it :)

PR #191, opened 2023-10-19
Add "speedrun" timer mode that includes skipped cutscenes and doors

NobodyNada:
Removes the "Adjust Fanfare Timer" option. In its place, the "Realtime/Ingame" timer selector gets a third option, called "Speedrun" (could be renamed to "Vanilla" or "Adjusted" or something). This option is real-time except as it would be during a run of the vanilla game, so skipped item fanfares & boss intros, as well as fast doors and elevators, have the skipped time added back in to the room and segment timers.

Upon upgrade, users who had the "adjust fanfare" option turned on and their timer set to realtime will automatically have their timer changed to speedrun mode.

### DIFF
--- a/src/fanfare.asm
+++ b/src/fanfare.asm
@@ -180,28 +180,11 @@ hook_message_box_wait:
     DEX
     BNE .nofanfareloop
 
-    LDA !sram_fanfare : BIT !FANFARE_ADJUST_REALTIME_8BIT : BEQ .done
     %a16()
-    LDA !ram_realtime_room : CLC : ADC #$0148 : STA !ram_realtime_room
-
-    ; adding 5:28 to seg timer
-    STZ $12
-    LDA !ram_seg_rt_frames : CLC : ADC #$001C : STA !ram_seg_rt_frames
-    CMP #$003C : BMI .add_seconds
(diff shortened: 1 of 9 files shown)


src/fanfare.asm
```diff
@@ -180,28 +180,11 @@ hook_message_box_wait:
...
-    CMP #$003C : BMI .add_minutes
-    SEC : SBC #$003C : STA !ram_seg_rt_seconds : INC $12
-
-  .add_minutes
-    LDA $12 : BEQ .done
-    CLC : ADC !ram_seg_rt_minutes : STA !ram_seg_rt_minutes
-
-  .done
+    LDY #328
+    JSL ih_adjust_realtime
```
InsaneFirebat, line 185: We later realized that this routine runs during all message boxes, including refill stations. To fix it, I moved the code to `ih_get_item_code` in infohud.asm, but I hadn't made a pull with that fix yet.

InsaneFirebat: The Y register needs to be preserved before adjusting the timer. It points to the item PLM's instruction list, which is read shortly after `ih_get_item_code` returns.

InsaneFirebat:
Merged with #193

PR #192, opened 2023-11-05
Some minor commits and fixes

InsaneFirebat:
Not much going on lately. Here are some minor things that have come up since the last update.

- Instead of `LDA #$2804 ; D`, we can write `LDA #$2800|'D'` and the value for D ($04) will always be updated automatically.
- 14% presets get max PBs for the Ridley fights.
- Combined some of the crash handler code and fixed a dumb thing I did (a lot) that asar was kindly ignoring.
- Fixed uninitialized controls after teleporting or loading a preset before selecting a save file.
- New menu palette profiles from MemesAreHealthy and grapedrinkz.
- Added the G4 skip code to the menu toggle so it works if toggled while already in G4 Hallway.
- Automatically update timers when Crocomire changes music.
- Fixed some unreachable code due to flags not updating on STA instructions
- Swapped out a push/pull for faster register transfer instructions


InsaneFirebat:
Merged with #193

PR #198, opened 2024-02-12
Skip HDMA handler and infohud game loop when gameplay is skipped

idlechild:


idlechild:
Finished testing moondance strat with a TAS.  Tomorrow will do some manual testing, hopefully make a tutorial with it and then push it out.

I'm not seriously asking for a code review of the moondance strat.  Among other things it's comparing poses (and trusting the game to present those poses when expected) so almost guaranteed it's not 100% correct.  I'm already aware of some corner cases where the moonfall count can get off.  As long as it's close it's still helpful (and if you're doing the moonfalls correctly then the count should stay correct throughout).

If there are any other concerns with this PR, now's the time to bring it up.

src/defines.asm
```diff
@@ -685,6 +685,7 @@ endif
 !ACTION_MANAGE_PRESETS      = #$0028
 
 !TOP_DISPLAY_VANILLA = #$0002
+!TOP_DISPLAY_VANILLA_8BIT = #$02
```
InsaneFirebat, line 688: These _8BIT defines aren't necessary. They can be shortened with .b on the opcode

idlechild:
Never mind on that tutorial, I am bad at moondance LOL.  Also having difficulty finding a comfortable grip.  I do think the trainer is helpful though.  I'm just going to release it now.

PR #202, opened 2024-02-29
Optimize kraid rock projectiles to reduce lag when Kraid rises

idlechild:
Kraid rocks ended up being the answer; credit to metconst for suggesting it... although I was correct that their solution was overkill.  Their suggestion took the "laggier" fight and made it on par with the "laggy" fight.  I just created optimized versions of the routines, which was also overkill so then I only optimized one of the two routines.

I primarily tested with Samus in the KQK position but remaining morphed the whole time and then measured how many RTA frames it took to execute the 9 seconds of IGT from start of Kraid rising.  Despite this being a fairly controlled scenario, I found that varying the time I spent loitering in prior room affected lag by 1-4 frames.

2.5.12.1: "Laggier" Kraid is 139-143 lag frames, "Laggy" Kraid is 61-64 lag frames
2.5.12.2: "Laggier" Kraid is 125-126 lag frames, "Laggy" Kraid is 52-53 lag frames
Vanilla: "Laggier" Kraid is 123-125 lag frames, "Laggy" Kraid is 51-52 lag frames

I'm going to post ips files on the discord in case anyone wants to test it.

Oh also, despite using version 2.5.12.2, I don't mean for this to be a hotfix.  It can go in the next regular update along other changes.


idlechild:
Built something granular for clearing enemies.  No more kihunter wings.  Kept the bosses alive as that's usually desired (if you already killed the boss then clearing enemies has no effect anyway).  The choices for what to keep/what to clear might need to be tweaked but it's easy to do so.  I haven't tested PAL but I expect it to work there too.

idlechild:
Okay are we ready to release this?  What's the verdict on those glitch tiles?  I don't really care; anyone should be able to put their changes onto this PR to set it how they want.

NobodyNada:
I'm working on a patch for the glitched tiles right now.

idlechild:
Same for the release notes, can leave it with the todo, or go with the nintendo staple:
"General system improvements to enhance the user's experience."
Or perhaps something else

NobodyNada:
I feel like quickboot and Kraid lag reduction would benefit from real release notes, since those are changes that directly impact how people might choose to use the practice hack.

idlechild:
We can always add those later (probably update help later too), and I shared the Kraid stuff on discord already, and quick boot is defaulted off unless you have the dev build

idlechild:
I was originally thinking there'd be some toggle for the Kraid behavior, but it ended up not being that dramatic of a difference, and it's not like it was correct before.  I'm sure it's not correct now either; it's just closer.

I'm just keen to see these demos revealed by people discovering it.  Bumping up to 2.6.0 with no release notes could get the job done :)

Or maybe I'm alone in this.

IFB any thoughts?  What's a good time to drop this?  Or am I out-of-the-loop and you've already shared it with a bunch of people?

InsaneFirebat:
All good here. A few people know about them but it's generally still unknown. I was planning to leave it out of the update notes.

idlechild:
Alright I got outvoted on the vague release notes, now added

NobodyNada:
I think the "general improvements" one would be funny to add on the end, just to hint that there's something fun in there, but it's your call :)

PR #203, opened 2024-03-03
Use correct VRAM transfer mode for savestates

NobodyNada:
We set the high bit of [`$2115`](https://snes.nesdev.org/wiki/PPU_registers#VMAIN_-_Video_Port_Control_($2115_write)) to 0, which instructs the PPU to increment the VRAM word address after writing the low data byte. However, the DMA transfer [writes low bytes first, then high bytes](https://snes.nesdev.org/wiki/DMA_registers#DMAPn). This means the high byte of the first word of a transfer was being skipped, leading to an occasional glitchy tile. Setting the mode bit to 1 instead fixes the issue.


InsaneFirebat:
Nice. That's probably the bug total mentioned a long time ago but I couldn't make any sense of it at the time.

InsaneFirebat:
Merged with #202

PR #204, opened 2024-03-07
Custom title menu demos

InsaneFirebat:
I made some small changes to the way demos are loaded. The most important change is at the top of demos.asm, and allows the large chunk of input data to live in any WRAM bank ($80-BF). There's also a fix for lava/acid damage not being applied, and some slight optimizations for rom space so that very little freespace is needed outside of the big stuff. There's now a big chunk of freespace in bank $91, which starts at the end of `DemoRoomReset`.

The first six demos were reproduced on PAL, while NTSC had all 23 demos replaced. Not all of them are particularly interesting, so feel free to suggest replacements (for a future update). I created a macro to handle warnpcs when they differ between PAL and NTSC. Found a few existing places to use it and updated them.

### DIFF
--- a/src/cutscenes.asm
+++ b/src/cutscenes.asm
@@ -833,11 +833,7 @@ endif
     LDA !sram_cutscenes : BIT !CUTSCENE_FAST_MB : BNE .fast_mb
     JMP cutscenes_mb_normal_init
 }
-if !FEATURE_PAL
-warnpc $A98824
-else
-warnpc $A98814
-endif
+%warnpc($A98814, $A98824)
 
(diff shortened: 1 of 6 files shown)


idlechild:
I'm trusting this is awesome can't wait to see it!

I propose we up version to 2.6.0 with incredibly vague release notes.

InsaneFirebat:
Sounds good to me. I just pushed a fix that had the first few demos overwritten by something else. Moved it to bank $87

InsaneFirebat:
Merged with #202

PR #206, opened 2024-03-24
Add PAL Debug option to NTSC

idlechild:
PAL Debug option is also persistent in SRAM, mainly because it was easier to code that way.
Also ported some improvements from IFB's HackROMBase branch.

I want to add tasdance support to moondance room strat, and then there is a request to improve charge timer.  There's also a request to add HBJ trainer which I am ignoring (sorry!)  There was a request to fix something about the menu shortcut; do we have a fix for that already?

Is there more we wanted in the next update?  I feel like I'm forgetting about something.


InsaneFirebat:
> There was a request to fix something about the menu shortcut; do we have a fix for that already?

I didn't write anything at the time. You could move the menu shortcut check to the top, right before save/loadstate, in gamemode.asm. Or you could put this untested code at line 1714 of menu.asm. (or both?)
```
    ; disallow inputs that match the menu shortcut
    LDA [!DP_CtrlInput] : CMP.w #!sram_ctrl_menu : BEQ .store
    CMP !sram_ctrl_menu : BNE .store
    %sfxfail()
    ; write code to set cursor position to 0 (menu shortcut)
    LDA #$0000 : STA !ram_cm_stack_index : STA !ram_cm_ctrl_last_input

  .store
    ; Store controller input to SRAM

File: src/infohud.asm [hunk 1]
```diff
@@ -612,7 +612,12 @@ ih_update_hud_code:
     ; Map visible, so draw map counter over item%
     LDA !sram_top_display_mode : CMP !TOP_DISPLAY_VANILLA : BEQ .mmVanilla
     LDA !ram_map_counter : LDX #$0014 : JSR Draw3
-    LDA !sram_display_mode : CMP #!IH_MODE_SHINETUNE_INDEX : BNE .mmRoomTimer
+
+    ; Skip room timer when shinetune or walljump enabled
+    LDA !sram_display_mode : CMP #!IH_MODE_SHINETUNE_INDEX : BEQ .mmSkipRoomTimer
+    CMP #!IH_MODE_WALLJUMP_INDEX : BNE .mmRoomTimer
```
InsaneFirebat, line 618: Do you think this is worth having a flag that the HUD mode sets and this routine clears?

idlechild:
Alright that's the last of the infohud modes that needed some improvements.  Sheesh that shinetune bug in the last number has been out there for a while and ended up being simple to account for.  I'm not sure it's perfect but it's definitely improved.

So it's just tasdance support.  As soon as that's done (or I give up on it), then I think 2.6.1 will be ready to ship.  Any objections?

idlechild:
I'm going to test it a bit more in an hour, but it is otherwise ready for release.  I'll likely release it later tonight.

InsaneFirebat:
I haven't caught up with the last few commits yet but I'm good with it either way

PR #210, opened 2024-04-10
Door HUD mode

### LINKED ISSUE #213: Metronome; cannot change the frequency
I cannot change the metronome frequency; it defaults to 10, and any DPAD press just beeps, and it flips between 1 and 255.
This is with a rom patched for everdrive X6, generated via the website on June 26th if that leads to any additional information.

InsaneFirebat:
If the user has Enemy HP set as their InfoHUD mode, then we can safely overwrite it during a door transition. I'm using this to draw my horizontal speed, but it also supports a few other HUD modes that only draw in those four tiles. Some of them were duplicated and optimized for one-time use (no checks if it's already been drawn) or to only draw in the designated area.

There are also some new menu palette profiles from PapaSchmo and Vespher.


src/infohud.asm
```diff
@@ -430,25 +430,44 @@ ih_before_room_transition:
...
+    PLB
     CLC ; overwritten code
     RTL
 
   .vanillaDoorLag
-    LDA !ram_minimap : BNE .doneDoorLag
+    LDA !ram_minimap : BNE .done
     TYA : JSR Draw2
-    BRA .doneDoorLag
+    BRA .done
```
InsaneFirebat, line 460: These branches should go back up to line 442. Not sure why I moved them to done.

idlechild (commented):
Nice idea, looks good

InsaneFirebat:
Added a fix for our IGT room timer not being incremented when Samus is locked in place

idlechild:
If %a8() isn't set before the math, then the math relies on the high byte being zero.  I ran into this issue in arcade; maybe it's not an issue on the vanilla practice rom?  At least not yet?  I'd rather it not rely on it though.

Maybe the correct solution is what you have but also do an AND #$00FF before the math.

InsaneFirebat:
For some reason, I was thinking there was already an AND #$00FF at the top of the routine, but the value isn't loaded until after the direction is determined. I also missed the two other places where the value is stored back to RAM. I think it should be safe now.

PR #215, opened 2024-08-17
Fix default controller bindings when booting from blank SRAM

NobodyNada:
The call to init_suit_properties_ram overwrote the carry flag


idlechild:
Cool, cherry-picked into my presets PR

PR #217, opened 2024-11-17
Practice Hack 2.6.4

idlechild:
- Elevator Entry option (short, long, off, defaults to short)
- Ceres elevator category presets start with Samus entering the room (regardless of elevator option)
- Add Snes9x 1.61+ to top platform line, and VC replaces Older Emulators
- Add health alarm always on option
- Free up space in bank 80 and 82 by moving a bunch of preset logic to another bank (arbitrarily chose bank 83)

I used the enemies in green brinstar main shaft to calibrate the long option elevator.  I suspect it is not quite TAS/frame/pixel-perfect but it is close enough.  Long option elevator also has a hitch going down morph elevator that no other elevator has.

If you create custom presets while standing on the elevator (or force-stand in any room with an elevator) then you get the elevator behavior.

Only thing left to do is review the category presets, perhaps adjusting if the preset is before or after elevator, or if there should be one on each side).  I'm going to leave this up for review until Sunday afternoon or evening and then push.

Any objections to the platform name changes?  Is Snes9x 1.61 the right version where 256k SRAM support added?


idlechild:
Shoot I just noticed the maridia elevator room background is messed up with a short entry down the elevator (obvious once you start falling through the room).  I'll want to fix that before release.

PR #218, opened 2024-11-20
Port many improvements from IFB's code review

idlechild:
I'm not in a rush to make another release unless people are screaming for spin lock (with current tourney being vanilla I wouldn't expect it).  I upped the version 2.6.4.1, but I assume this will be included in some eventual 2.6.5 update.  I'll also feel more comfortable after it's gotten more testing.

CPU report:
- In $9181A9 method, replaced JSR $81F4 with JMP $81F4, saving three cycles
- Moved the rest of the $9181A9 method to the end of $81F4 method to get rid of the RTS, saving six cycles
- Added LDA !sram_spin_lock : BEQ .resume_vanilla, spending nine cycles to come out even if spin lock is off

I had to move a couple methods and one table around in bank 91 to make it work.  Had a funny bug where I didn't notice I broke the branch at the very beginning of $9181A9 method, but it happened to not crash the game, but also not work right.  You'd start running with Samus, let go of controller, and she would keep moving lol.  That bug is fixed now.


idlechild:
PR previously was for adding spin lock option, but the IFB changes are more significant so renamed the PR.

For the most part focused on aligning code on existing features rather than porting over new features.  That said there were some features and improvements ported:
- Six preset categories added
- Menu supports slowdown mode
- Timer values retained while menu is open
- Option to control big baby metroid RNG
- Option to play fanfare music even when other music is off
- In speedrun timer mode, do not add time if fanfare enabled
- Second controller L/R couldn't select RAM watch
- Do not rerandomize RNG in dust torizo and big baby rooms
- Added EXAKTScience profile
- Added option to build practice hack with vanilla HUD
- Added WRAM_START to ROM at 80:FFF0 for cout's tracking tool

I also redid the TODO file to list the github commit last used for porting, as well as listing some features we may want to port, and also some differences currently not planned to be ported or resolved.

idlechild:
There were some design/coding style changes ported in as well, some new while others a continuation of previous work.  I wanted to briefly cover it here:

- More labels for vanilla WRAM.  I believe IFB was the one who started that tradition, and he added more labels.  I also added more as I was porting this in, particularly in features not in the IFB branch.  There's still more that could be added.

- INX #2 format of listing multiple repetitive commands (NOP #2, INC #3, etc.)  I was previously against this as I thought developers might think they are single instructions.  I've since come around and acknowledged it's something assembly developers need to know.  Note that IFB did favor replacing some of the longer ones with math, for example replacing INC #4 with CLC : ADC #$0004 since that is faster.  I saw at least one BRA $00 replacing a NOP #2.  I also thought this was good practice going forward.

- Using .l or .w or .b on commands using assembly labels, and not using those for commands using constants (or defines to constants).  This makes sense considering the usage for assembly labels is ambiguous while it is not ambiguous with constants.  IFB also favored using hex constants instead of decimal constants since you can explicitly specify the size (long, word, byte) of hex constants.

- Replacing BRL with JMP.  My mistake for not realizing BRL is worthless.  I thought it was the same as JMP but it's actually inferior.

One change I did not port over was replacing TDC with LDA #$0000.  In fact I doubled-down and searched for more instances of LDA #$0000 that could be converted to TDC (except in the crash handler; in a crash scenario it's possible a TCD instruction got executed so it is better to use LDA #$0000 instead of relying on TDC there).  When you see all the new uses of TDC in this latest commit, that came from me not IFB.  Yes this means the two codebases will remain somewhat different.  There are some other differences called out in the TODO text file as well.

InsaneFirebat:
If you had Vanilla HUD and minimap enabled, an infinite loop would occur when touching a door transition, due to the Draw2/3 calls being skipped which would have returned the A register to 16bit. This is fixed now.

idlechild:
I think this will be ready to release today.

I still want to port IFB's SuperHUD, and there's also a map rando request for customizing door locks.  I think those will take longer so they can go in the next release.  I feel like I'm forgetting something but if so that could also go in the next release.

I'm going to test some more before release.  Let me know if there's a reason we shouldn't release today.

PR #220, opened 2025-02-08
Implement cout's freespace macros

InsaneFirebat:
Replaces our orgs with a macro that assigns freespace to the next available byte. Use `%startfree(80)` to org into the next byte of freespace in bank $80. Use `%endfree(80)` to signal the end of freespace usage and update macro variables. Replaces most of the print commands with a list of banks written to, the next available byte, and how many bytes are remaining.

From https://github.com/cout/baby_metroid/blob/main/src/freespace.asm


InsaneFirebat:
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

This sounds like it needs @NobodyNada attention

InsaneFirebat:
Merged into #223

PR #223, opened 2025-02-24
Porting all the HUD modes

idlechild:
Super HUD is here!  I expect some corner cases won't work, but it seems like it works well enough.  It could use some more testing, but I am also planning to port in some more from IFB branch (at least the BRB menu) so probably won't beat up on it too much until that's done.  Also the Door HUD feature isn't working but I suspect it was broken before I started... but also perhaps by reviewing differences between IFB branch and our branch one last time I can locate what I am missing.  Maybe the reason for the separate infohud routines that IFB had and I didn't like has to do with Door HUD?

Also need to integrate Nada's Phantoon PR and IFB's org PR.  I might actually need the org PR for the infohud as it is almost taking a whole bank now, although both custom number gfx and spriteprio could be moved out if needed.


InsaneFirebat:
Door HUD is breaking because we're overwriting $12 in some of the routines. We should probably get rid of our remaining $12/$14 usage. We could swap the Draw routines to use $3E..43. I think PJ reviewed the DP usage not too long ago. That one might not have been there last time we looked for free DP.

idlechild:
nothing to see here >_>

NobodyNada:
Just some general system improvements to enhance the user's experience

idlechild:
Infi-Doppler seems to be working.  I noticed it was easy for the mode to be left on, so I added some sanity checks so it shuts itself off.  I also realized Infi-Doppler can use RAM near the start of bank 7F (just past the first screen of data) since we know Phantoon has a small room.

I'm sure we could find a better way to organize the RAM, and I'd still prefer the practice hack not hit the SRAM as often as it does, but this would be a fair amount of work.  Not taking that on right now.

InsaneFirebat:
> I also realized Infi-Doppler can use RAM near the start of bank 7F (just past the first screen of data) since we know Phantoon has a small room.
This seems ok as long as it's not using any of that RAM when left off. Most other rooms would wipe it clean and I don't expect anyone but Ruya to go straight from 1 round practice to OoB travel.

idlechild:
This is about ready for release.  It's not perfect (looking at you door mode) but I think it is a good improvement.  I may ship it tonight.

Btw thx IFB for the tip about our $12-$18 usage, that would be risky to fix those routines but maybe we could have door mode use their own routines or something.

TODO updated.  I also want to redo controller shortcuts to support options with the second controller, but that will take more time.

Note I'm leaving for vacation tomorrow afternoon.  If there's some dumpster fire of a problem, anyone around this week to fix it?  I'll be checking discord but will be limited on what I could fix remotely.

InsaneFirebat:
I'll be around but I probably won't know unless someone messages me.

idlechild:
I'm updating change log and shipping this in about 15 minutes

PR #225, opened 2025-03-09
Port remaining features from IFB practice hack

idlechild:
Made some improvements along the way.  Reserve/vanilla+ HUD mode is fixed.

I still need to figure out the door HUD display, and I want to do some more testing, but in theory this is ready.  I'm thinking about releasing it tomorrow.  RIP controller 2 shortcut support that I said would be in next release; hopefully the release after this one.


idlechild:
Okay, actually approaching release now, hopefully later today.

For some reason I thought importing IFB custom presets was incompatible but actually they are compatible, so that was nice.  You can use the factory reset option to keep those while resetting everything else.

PR #227, opened 2025-03-25
Improve Pseudo G-Mode

InsaneFirebat:
Toggle palette FX and enemy projectiles with pseudo g-mode
Fix misplaced infidoppler define
Shift temp menu RAM +$10 bytes to give us more space for permanent menu RAM


idlechild (commented):
I'm going to trust this is an improvement for G-Mode.  Looks good anyway.
Thanks for catching my mistake in the defines.asm; I noticed it a few days ago when working on controller shortcuts.

I will incorporate this into controller shortcuts PR, which is Coming Soon™

PR #228, opened 2025-03-25
Controller Shortcuts are now Awesome

idlechild:
For next release, planning to bump the version to 2.7.0.  While the initial out-of-the-box shortcuts won't really change, it's quite the change to how they work internally as well as the available options.  You now have control over shortcuts for both controllers (previously all controller 2 shortcuts were hard-coded).  You can choose if shortcuts require exact match or not.  You can map the same inputs to multiple shortcuts so you can cause multiple actions with the same input.  Also the practice hack menu accepts inputs from either controller.  I plan to be putting this out for beta test in the near future.

It is also technically a compatibility break?  I could write logic to retain the user's previous shortcuts but I don't see the point in doing so.  I think most users either won't care (stick with the default shortcuts) or they will be interested in redoing their shortcuts.

There are still several things to sort out:
- I need to protect the WRAM through spacetime beam (and also xray shenanigans?  not sure if xray can modify the end of bank 7E)
- Pause and unpause don't work correctly anymore
- Factory reset removes the save and load shortcuts (it also crashes because of sound but I suspect it is the same issue of clearing WRAM that perhaps shouldn't be cleared... although the first quick fix I tried didn't work so might be more to it)
- You can no longer load state out of a fanfare or door transition, but it'd be nice if you could like before.  If you crash, you can only load state using the hard-coded default of Select+Y+L and I think that is okay in the crash scenario, but I'd prefer to use the user preferences for fanfare or door transition scenarios.

If you have questions about the design, please refer to the official design objectives for this feature:
Design Objectives


idlechild:
I think it's ready to beta test but too late in the evening to make a post about it, can do it tomorrow.  Also the splash screen isn't much of a splash screen, but at least it is something.  Can improve it before 2.7 release.

InsaneFirebat:
Some feedback after testing:
- The timeout for a shortcut going blank seems too short. It's caught me a couple times and closed the whole menu (because my shortcut contains Start or Select), which deletes the shortcut from the list.
- Can the shortcut be ready to set when we add it to the menu? That also caught me off guard a few times and closed the menu.
- Soft Reset should go back to its vanilla behavior. It doesn't work in a lot of places it used to. Especially the menu which shares half of its button combo with reset.
- It would be nice if we could make "Add Shortcut" stand out a bit more in the menu. With all the default shortcuts in the list, it feels kinda buried.
- We used to be able to delete a shortcut by pressing X on it. It wasn't intuitive, but it was way faster than deleting shortcuts by timing them out (especially if the timeout is increased).

InsaneFirebat:
Found a problem with the option to skip the game over menu. If there's no savestate to load, it loads the last preset pointer and set it as the preset to load. The code that checks for a preset to load is no longer being run unless a controller shortcut triggers it.

idlechild:
Thanks for the feedback!  Not ignoring it, just haven't done anything with the practice hack in over a week.  Hoping to pick it back up this weekend.

InsaneFirebat:
I know that feeling. I just took a few weeks away from the disassembly.

idlechild:
I still haven't addressed IFB comments and still more to do, but at least I've returned to development.

idlechild:
Addressed some of IFB's comments.  I'm sure the Add Shortcut option could be better, but it is working well enough.  I also prefer removing the hard-coded soft reset and making that customizable.  We are running the controller shortcut routine in more situations now so it's not as much of a loss not having the hard-coded reset... but it is still a potential loss if you used that; you can't reset out of the practice hack menu and some other situations where the game is busy.

There's enough different that we may get some feedback when 2.7 releases, but I think it's ready.  I have one more map rando related thing I want to try to get done this weekend, and also I'll likely update a few left/right portal names if the map rando community prefers it, and then I think this will going out the door; should be in a few days or so.

PR #250, opened 2026-05-03
Allow customizing streamer name in BRB menu

InsaneFirebat:
The bulk of the pull is for the BRB menu. The keyboard has been repurposed again to let the user choose the name that appears on the top line of the BRB screen. The name gets auto-centered horizontally with a bias toward the right.

The artificial lag setting that almost no one should be messing with was moved from the InfoHUD menu into Slowdown. Not saying that's a good place for it, but it's better than it was.

Renamed the Compressed option for preset to Custom Samus Skin.

There were drop tables at the very beginning of mainmenu.asm. I assume this was so the define was near the table and defined before it's used. Calculating the table size automatically lets it be hidden away in defines.asm where it will still be defined early enough.

bastion_blue_succubus87 sent me her menu palette profile forever ago and I'm just now getting it in.


idlechild (approved):
Looks good.  I'm going to use this as the base for a pull request (I'm also working on some stuff for the next update)

PR #251, opened 2026-05-03
Release 2.7.4

idlechild:
Builds on top of IFB's pull request.  I was almost ready to make my own PR, so good timing can just include my changes on top.  I still need to update the help pages.  I'm also going to be making a beta test build from this PR; I'm not sure if the map rando stuff is all correct.


InsaneFirebat:
Be sure to grab the latest commit from that pull to fix a bug or two and init the SRAM

idlechild:
My first attempt at maintaining persistent WRAM through soft reset failed because the WRAM wasn't getting cleared on hard reset.  The hard reset and soft reset code flow is weird; I suspect it is "not correct" but it does work so I guess that's the important thing.

What's weird is, when you look at Bank 80 disassembly, you'd expect the $841C-$8461 only executes on hard reset, and $8462-$8481 only executes on soft reset, but that's not what happens.  On hard reset $841C is executed first but $8462 will be executed later (maybe correct?), and on soft reset $8462 is executed first but $841C is executed later (probably not correct!)  However since we've taken over the audio handling and maintain !ram_quickboot_spc_state throughout, we don't end up messing up the audio.

We can use this flag for initializing the persistent WRAM as well.  If it's zero then we can assume nearby persistent WRAM is also either zero or properly initialized (likely a soft reset happened), so no need to initialize persistent WRAM.  If it's not zero then we probably need to clear all the other persistent WRAM (likely a hard reset happened).

Now I just need to look into RAM watch.  Request was to maintain minimap setting through soft reset (done) and also was pointed out RAM watch settings don't carry over (they still don't, will look into that).

idlechild:
I still need to test the phantoon logic but at least the menu is functional

idlechild:
Alright what's left for release 2.7.4?

I know I still need to update help and release notes.  Also someone brought up the inconsistent WRAM implementation (persistent WRAM doesn't work like it was originally intended) so I plan to work on that.  That's all I have on my end.

@tewtal Seems no one is clamoring for the InfoHUD palette change; can I assume we're leaving that on the back burner?

@InsaneFirebat You had a different idea for customize menu; how's that coming along?

PR #254, opened 2026-05-21
Make OOB Tile Viewer compatible with any% glitched

idlechild:
It's not quite there yet.  Gold 5D blocks no longer crash, and I'm able to collect X-ray.  However:
- The attempt to clean up OOB Tile Viewer sprites during block shuffler does not work.  It messes up the setup for first 6F block.
- The TAS movement to the door transition stopped working.  This could be just an error on my part; I might have modified that section by mistake.  I still need to review it.

Also added an any% glitched submenu under InfoHUD -> Customize RAM Watch -> Common Addresses.


idlechild:
Oof make a couple silly mistakes in this PR, no wonder OOB didn't work right after block shuffler.

I think the concept works though; I just need to clean it up.

PR #258, opened 2026-05-25
Improve OOB tile viewer

idlechild:
I'll try attaching an image of how a section of parlor used to look.  The 1x2 and 2x2 bomb blocks do not all show up.  Additionally scroll blocks (which are air) appear as blue blocks.  While I understand why these special air blocks were being drawn (they're sometimes relevant, although I'm going to draw them as pink instead of blue), I don't want to see scroll blocks, so I'm excluding those.  These changes also fix door caps (all shot blocks appear instead of just the primary shot block), and they also help with the any% glitched route.  There's a bomb block (via vertical extension) that we jump on that previously didn't appear on the OOB tile viewer but now it does.  Also fixed a glitch with the RAM watch menu (we should default to bank 7E not 00).

image

### DIFF
--- a/src/defines.asm
+++ b/src/defines.asm
@@ -1479,21 +1479,22 @@ endif
 ; Each botwoon pattern can be #$00 (off), #$01, #$09, #$11, #$19
 ; (except hidden pattern cannot be #$19)
 ; This corresponds to just three bits of information
-; For efficiency this information can be overlapped,
+; For efficiency this information can be overlapped
+; and is done for the hidden and second patterns,
 ; allowing for the spit value (#$00, #$04, #$08) to be included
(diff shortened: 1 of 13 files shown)


idlechild:
Working towards a 2.7.5 release.  Included a few more changes.

I think I need to change RAM watch for the 00-2F DP variables.  I keep getting tripped up that the 0026 and 0028 variables I am viewing in the RAM watch menu are not the real variables.

idlechild:
Added logic to automatically redirect DP register access to the DP register backup while in the RAM watch menu.  I still need to verify it works, but this is what I'm trying to do.  It won't handle the edge cases of 16-bit read/write from FFFF or 002F but if it works in the more common use cases then that's good enough.

PR #262, opened 2026-06-11
Release 2.7.5.3

idlechild:
Received a couple of simple requests.  Also including a RAM watch common address.  I also want to adjust the door skip infohud mode to show the $1842 value when you hit the block shuffler.  Just need to do that and some testing, hoping to release later today.


idlechild:
I thought I broke vertical speed when shinesparking up or diagonal, but turns out that value is just messed up vanilla.

I did break the rando mother brain presets so will be making a fix there.

PR #268, opened 2026-07-09
Add option to show extra OOB tiles where the X position wraps around

idlechild:
This feature turned out better than expected.  I might push it tonight.  Also fixed a bug in my x-ray logic.

I do want to add a 0% RAM watch menu, and also finish working on 0% TAS.  Not sure if I can get that done tonight or not (the hard part of the TAS is done though).


idlechild:
Planning to release sometime tomorrow (or today depending on when you read this)

PR #269, opened 2026-07-11
Sprite features now has option to show hitboxes in blue or white color

idlechild:
Going to get feedback on this first, but seems simple enough


InsaneFirebat:
Iirc, grapple usage will overwrite the palette for non-tile hitboxes. Can't check at the moment and will probably forget to later.

idlechild:
You're right.  So that's what causes the color to be light blue (it's light blue in my any% glitched TAS sometimes and I hadn't gone back to figure out why).

I wonder why it's not light blue all the time... maybe we could make that an option as well.

I'll have to see where grapple is doing that.

InsaneFirebat:
> I'll have to see where grapple is doing that.

https://patrickjohnston.org/bank/9B?just=C689&highlight#fC689
Without it, grapple beam is dark blue

idlechild:
I'm aware this subtly changes Ceres elevator appearance when Samus enters/rides the elevator (noticeable if you change the hitbox color) but I think it is acceptable.

PR #271, opened 2026-08-02
Add map rando room layout option

idlechild:
Map rando option also patches the sound queue bug that causes blue echoes to be lost in heated rooms, and it enables "aim anywhere" (i.e. you can set angle down to select or a face button).

TIL the vanilla game always does this.  When you put dash on L, it puts angle down on select.  It then has some menu logic to cover that up with "OFF", and then within the game it explicitly checks for L or R for angle up and angle down and will discard the input if it is not coming from L or R.  This explains why when I go into prac hack menu -> game options -> controller setting mode, it would show select assigned to angle down since the binding is there.  It's also somewhat important to the vanilla game that it works this way; it allows the vanilla menu logic to work by always swapping inputs, since all seven controller input options are always assigned to something so swap is always valid.

To enable aim anywhere without changing the CPU cost of the in-game checks, I'm changing how the bindings are loaded.  The in-game check will now permit any angle up or angle down input (it still checks that the input is valid but the input will always be valid).  To make it still work similar to vanilla when map rando option is off, I'm now unbinding angle up and angle down if they are invalid.  I can't do this all the time or I expect it would break the vanilla menu, so I'm only doing it when you enter gameplay, and I have to put the bindings back whenever you save to SRAM.  I also have to put the bindings back when entering prac hack menu and leaving prac hack menu.  Finally I needed to modify prac hack menu to expect the incorrect bindings to be present and hide them when not in map rando mode (which arguably should have been done anyway; as mentioned above prac hack would show angle down bound to select even when that binding didn't work).  Prac hack controller setting mode will also allow any input to be selected if map rando option is enabled.

It's not perfect in the case where both angle up and angle down are assigned to select and/or face buttons, but I can't think of a real world situation where you would do this; I think it's good enough.

Not ready to pull this in yet.  I believe both IFB and Nada have some unrelated updates we can include.  Also I want to look into Update Timers.  If you set it to update every frame then it causes some lag, which we expect but it might be more lag than expected.  Also there was a request to make it better for Ridley room and since minimap is off during Ridley fight that could be more effort.


idlechild:
Sheesh Update Timers is doing a lot of work.  I forgot we update multiple values.  There's some extra work it was doing that we need when pausing/unpausing or opening/closing prac hack menu, but we don't need when calling Update Timers.

I made a separate, optimized method in bank 80 just for Update Timers.  No JSR Draw instructions, no $12 variable usage, doesn't even use Y register.  Even with this, we might still want to adjust artificial lag if updating timers every frame, and perhaps do something about Ridley.  More to investigate but posting the progress so far.

InsaneFirebat:
A HUD mode might have been a better approach to updating timers every frame. Maybe too deep now.

PR #272, opened 2026-08-02
Make some RNG calls after rerandomizing

jnkel:
This PR adds 8 RNG calls upon loading a savestate with rerandomize enabled. This is to try to break up patterns in case the game is in a weird RNG state caused by being in an XBA room that is doing a few predictable RNG calls every frame.

This intentionally has no effect on RNG loops outside of XBA rooms. Getting stuck in one of those loops involves visiting an XBA room for a specific period of time to move out of the main loop another RNG loop. Once in such a loop, further RNG calls will remain in the loop until visiting a room with XBA or with an enemy that reseeds RNG. We leave these alone in case someone is practicing a manip that involves intentionally getting stuck in an RNG loop. The type of RNG states this is intended to address are more transient, less-understood, and only exist while actively inside an XBA room.

Added because Zeni thought Ridley was having a tendency to get stuck in weird patterns even with rerandomize turned on; he has been testing this patch for a bit and hasn't reported seeing this behavior since. It may have just been confirmation bias, but if it is a real effect the only possible explanation I can think of is weird XBA states. So this patch might not actually do anything, but it can't *hurt*.

### DIFF
--- a/src/save.asm
+++ b/src/save.asm
@@ -109,6 +109,13 @@ post_load_state:
     JSL MenuRNG ; rerandomize hack RNG
 
   .randomizeOnLoad
+    ; Make some RNG calls in case we're in a weird XBA state.
+    LDX #$0008
+  .rerandomize_loop
+    JSL $808111
+    DEX
+    BNE .rerandomize_loop
+
     ; Randomize energy/ammo?
     LDA !sram_loadstate_rando_enable : BEQ .init_wram
     JSL RandomizeOnLoad
--- a/src/tinystates.asm
+++ b/src/tinystates.asm
@@ -148,6 +148,13 @@ post_load_state:
     JSL MenuRNG ; rerandomize hack RNG
 
   .randomizeOnLoad
+    LDX #$0008
+    ; Make some RNG calls in case we're in a weird XBA state.
+  .rerandomize_loop
+    JSL $808111
+    DEX
+    BNE .rerandomize_loop
+
     ; Randomize energy/ammo?
     LDA !sram_loadstate_rando_enable : BEQ .init_wram
     JSL RandomizeOnLoad


InsaneFirebat:
By my calculations, this routine does the same operations in 133 fewer CPU cycles. The main optimization is eliminating JSL/RTL, but it also uses temp RAM $C1..C4 to squeeze out a little more. Can live anywhere, but I'd probably drop it in the misc.asm pile.

https://gist.github.com/InsaneFirebat/ad4d781691a26b54cd67a8c4a79dfb62

Immediately after posting, I realized the first LDA $C1 is redundant. That would bump it up to 157 fewer cycles.

jnkel:
Nice, though I'm not sure saving 133 cycles on a loadstate is worth it. IMO simplicity / readability / maintainability is a more important concern here.

InsaneFirebat:
I get that logic, but JSL CycleRNGbyX can't be any less readable than JSL $808111 : DEX : BNE

jnkel:
For me, it's more about having that code live somewhere in another file, and having to review, understand, and maintain it on the future. But mainly, it's that I'm traveling this week and don't have the capacity to integrate and test those changes.

If you want to go ahead and make those changes and open a new PR, feel free, I won't object. I'm just not sure that the benefit is worth that effort.

InsaneFirebat:
I let myself get distracted by optimization opportunities. I actually came to see how you were detecting XBA rooms, but it looks like you're not. I'm not sure how this solves the problem of being stuck in a tiny predictable loop. I would also expect an odd number of cycles to be desirable, but I don't really follow the statistics behind this stuff.

Edit:  finished discussion in MetConst. tldr is number of cycles between XBA's is what's changing.
https://discord.com/channels/127475613073145858/370769325851803648/1534004254711222272

PR #274, opened 2026-08-09
Emu crash and tweaks

InsaneFirebat:
The only user-facing change in this pull should be in the crash handler. The rest is a bunch of things I wanted to do while reviewing code.

The crash handler now supports crashes in emulation mode and will print `E` above the stack dump. NMI is disabled ASAP, although this will likely have little to no effect based on how NMI is handled during a crash. I removed the call to vanilla code in the crash handler's CGRAM transfer, and adjusted some BRB/crop code that called its routine.

InfoHUD modes, room strats, and SuperHUD menu options now have their indices updated automatically. The tables in `infohudmodes.asm` are a little uglier, but it's way easier to rearrange or add new entries. `!BRB_TOTAL_SCREENS` was also automated. Remember to use .w/.b with these defines, since asar has to calculate the result and we can't rely on it assembling to 16bit anymore.

Use a lookup table instead of compare ladder on Zeb and Metroid event numfields.

Added a macro for the music toggle code in the menu. It's duplicated in another bank and this will make sure it's always in sync if it changes.

Replaced old compressed data bitmasks since they're no longer individually toggleable.

Found a few more places to use the PEI instruction. Also got reminded that it always pushes 16bits, like PEA, so don't try to use it in 8bit mode.


InsaneFirebat:
`../src/infohud.asm:2438: error: (E5088): Define 'IH_MODE_ROOMSTRAT_INDEX' wasn't found.`
This error is cursed... The file ends at line 2283. There are no references to `!IH_MODE_ROOMSTRAT_INDEX` in the repo. The hack builds fine locally.

I noticed in the Changes tab that line numbers aren't matching up with my local copy of `infohud.asm`. Did it not pull in all of my changes? Can someone do a test `build_dev.bat/sh` and see what's up? Maybe one of those python scripts I can't run is needed for something I changed.

idlechild:
The PR is auto-merging and then building.  If you pulled down canonical master then the 2438 line number would make sense.

InsaneFirebat:
Thanks, I'll take a look at that soon. I remember fetching on master but I might have started that branch a little earlier.

idlechild:
Not a fan of changing $80FFE4 to $00FFE4.  If it's still going to end up in bank 80 then I'd rather it say that in the asm.

InsaneFirebat:
> Not a fan of changing $80FFE4 to $00FFE4. If it's still going to end up in bank 80 then I'd rather it say that in the asm.

Vectors always go to bank $00, but I'll get it changed. I was about to make a new pull. You were right about the check failure.

PR #275, opened 2026-08-10
Tweaks and emulator mode crashes

InsaneFirebat:
The only user-facing change in this pull should be in the crash handler. The rest is a bunch of things I wanted to do while reviewing code.

The crash handler now supports crashes in emulation mode and will print `E` above the stack dump. NMI is disabled ASAP, although this will likely have little to no effect based on how NMI is handled during a crash. I removed the call to vanilla code in the crash handler's CGRAM transfer, and adjusted some BRB/crop code that called its routine.

InfoHUD modes, room strats, and SuperHUD menu options now have their indices updated automatically. The tables in `infohudmodes.asm` are a little uglier, but it's way easier to rearrange or add new entries. `!BRB_TOTAL_SCREENS` was also automated. Remember to use .w/.b with these defines, since asar has to calculate the result and we can't rely on it assembling to 16bit anymore.

Use a lookup table instead of compare ladder on Zeb and Metroid event numfields.

Added a macro for the music toggle code in the menu. It's duplicated in another bank and this will make sure it's always in sync if it changes.

Replaced old compressed data bitmasks since they're no longer individually toggleable.

Found a few more places to use the PEI instruction. Also got reminded that it always pushes 16bits, like PEA, so don't try to use it in 8bit mode.

This is a re-pull of #274 with a fix for my stale branch and initial feedback.


InsaneFirebat:
I meant to revert the change in `damage.asm`. That's every frame code and I didn't feel like balancing it, but you could if you wanted that cycle.

idlechild:
You're talking about the NOP : NOP replaced by BRA $00, yeah I saw that and debated if I should bring it up... it's barely a change but it's also an unnecessary change...

I would be in favor of NOP : NOP just to not change it but I'm not that opposed to BRA $00 there either.

Otherwise the PR looks good (at least what I saw initially; I assume there wasn't that much to fix after merging down master)

InsaneFirebat:
The merge was clean. Just had to fix the line that triggered the error and double check that there weren't anymore defines I hadn't seen. I'd like to get a little more testing in today, but this pull can sit on the back burner until there's something meaningful to include for players.

idlechild (approved):

InsaneFirebat:
~~I misclicked and pushed this stuff early. It's got a couple bugs I'm still working on.~~
Ready to go now.

InsaneFirebat:
As discussed with idle, the freespace prints have been expanded. It now prints used and total bytes per bank, and prints for all banks that have valid freespace. I had to modify create_names.py and create_clear_enemies_data.py to write manual orgs instead of freespace macros. Before and after examples below.

`Bank $85 ended at $85FA0D with $5F3 bytes remaining`
`Bank $85 ended at $85FA0D with $5A0D/$6000 bytes used, $5F3 bytes remaining`

PR #276, opened 2026-08-12
Avoid saving to SRAM when loading preset or savestate

idlechild:
Also fix a couple issues in the next-update branch.

Even though tile tables uses crossbank, the starting value of $E68800 was chosen so each tile table fits within a bank.  Starting at $E68000 was causing some issues (try jumping to LN or bubble mountain preset).  Also made freespace report the $800 bytes available at the start of bank $E6 (which I had forgotten existed, so nice to know about).

### DIFF
--- a/src/cutscenes.asm
+++ b/src/cutscenes.asm
@@ -1837,7 +1837,7 @@ else
 org $A98C05
 endif
     JMP cutscenes_mb_first_earthquake_start
-    BRA $02
+    BRA $02 : NOP #2
 cutscenes_mb_first_earthquake_end:
 
 if !FEATURE_PAL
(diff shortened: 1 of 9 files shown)


InsaneFirebat:
If the goal is to cut down on SRAM writes overall, I would axe `!sram_seed_x/y` entirely. It was a weird idea to keep the seed rolling forever since it only cycles when the menu is open, but can be used without the menu ever being opened.

We should add a comment on the tilegraphics org. I figured it was just a remnant of things moving over time.

idlechild:
Meh, it's fine to leave that.

This change came from a review I did of SRAM usage, where I was expecting to find we write to SRAM a lot but we actually weren't.  The write-on-load was the only SRAM write I thought was dumb and could be improved.

We read from SRAM a lot but it's not a problem to do that.

PR #278, opened 2026-08-13
Optimize and free up some menu memory

idlechild:
Freed about 750 bytes from bank 89, which was getting a little crowded but now has about 1200 bytes free.

About a third of it was just moved to another bank.  Most of the actual savings came from using X more efficiently.  For example:

TXA : CLC : ADC #$002E : TAX
...
STA !ram_tilemap_buffer,X
STA !ram_tilemap_buffer+2,X

Can be simplified to:

STA !ram_tilemap_buffer+$2E,X
STA !ram_tilemap_buffer+$30,X

### DIFF
--- a/src/BRBmenu.asm
+++ b/src/BRBmenu.asm
@@ -310,13 +310,13 @@ cm_tilemap_brb:
   .draw_timer
     ; Seconds
     LDA !ram_cm_brb_secs : ASL : TAX
-    LDA.l TimerNumberGFX1,X : STA !ram_tilemap_buffer+$360
-    LDA.l TimerNumberGFX2,X : STA !ram_tilemap_buffer+$362
+    LDA.l BRBTimerNumberGFX1,X : STA !ram_tilemap_buffer+$360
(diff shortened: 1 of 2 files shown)


InsaneFirebat:
The crash handler can be moved out of $89. It's $DB3 bytes on my end at the moment. Bank $B6 has $E00 remaining, and $AC and $B0 have $1200 if you want a little more breathing room.

idlechild:
Yeah next time it fills up we can move stuff around.

I want to add "safeties" to presets (idea from Lechtansi, who I think got it from Alttp).  Category adjustments was my attempt at a general solution, but a more specific safeties concept would be better.  I think it's doable.  I had some ideas in May and started on it, and then any% glitched took over.  Now I'm trying to remember what I was planning.  At least I had a start.

I don't think I'll need a lot of room in menu.asm for this, but if I do then we can move stuff around.

InsaneFirebat:
Bank $B8 is going to need some splitting soon too. It's all mainmenu.asm. SpritesMenu and TeleportMenu look like easy targets. Both of them could be dropped in a data bank with only the action_teleport's DB adjusted.

PR #279, opened 2026-08-15
Read-Only Locks for Savestates and Custom Presets

### LINKED ISSUE #208: Read-only custom presets
Add a way to lock a custom preset into read-only mode so it can't be overwritten.

InsaneFirebat:
Custom Presets can now be locked by pressing Y in the manage custom preset slots menu. This prevents them from being saved over, or deleted from the same menu. The savestate can be locked from the savestates menu, and prevents it from being overwritten, but not deleted by the dev mode option.

6 bytes of SRAM were allocated. The first 5 are for custom preset slots, and are bitpacked. The last byte is for the savestate. A new lock tile was added to menu graphics.

Closes #208


idlechild:
Seems reasonable.  I'm going to trust it works.

InsaneFirebat:
Yeah, this one is tested and ready to go

PR #280, opened 2026-08-18
Preset Safeties

idlechild:
Preset safeties allow a single set of category presets to be adjusted to be made safer/easier (or less safe/more difficult).  With this approach, instead of having five different KPDR presets, we should be able get down to two (late ice/early ice) yet have more flexibility than before.

This is a work in progress.  I'll likely force-push over this commit in the near future.  There's a lot of preset calibration still ahead.  I only have safeties added to KPDR 20% data (I haven't even renamed the preset files yet), and most safeties were added without consideration to how that impacts later presets (there's likely some strange behavior).  A couple safeties are not implemented yet, and I've only defined safeties for KPDR.

It's a start, and I thought it was worth raising a PR for awareness.


idlechild:
This idea isn't going away, but I don't think I need a PR just yet.

I mentioned trimming KPDR down to two preset categories with safeties, but perhaps going down to just one is possible.  For test purposes, I think I'm going to keep all the existing KPDR categories and add a new combined one for the safeties, with the goal of releasing with only the combined category.

For now though, closing this PR.

PR #281, opened 2026-08-18
Optimize Metroid and Zeb menu code (round 2)

InsaneFirebat:
Saved 62 bytes

I forgot to include docs for read-only locks, so that's here too.

### DIFF
--- a/src/flagmenu.asm
+++ b/src/flagmenu.asm
@@ -828,41 +828,16 @@ eventflags_setmapstations:
 
 eventflags_prepare_events_menu:
 {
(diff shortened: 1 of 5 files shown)


idlechild (approved):
The next update would be 2.7.10, but I'm hoping preset safeties is big enough to warrant a move to 2.8.0.

That said, if you want we can release 2.7.10 ahead of all that.

InsaneFirebat:
I'm in no rush, but it should be in a good place to merge. You can bump it again when the big stuff is ready.

PR #284, opened 2026-08-20
Write conditionals in symbols.asm

InsaneFirebat:
Updating the symbols script to also copy conditional directives lets us get rid of the duplicate defines for tinystates.

I edited python so someone (idle) needs to make sure I didn't do something stupid.


InsaneFirebat:
Merged with #285

PR #285, opened 2026-08-21
Add shortcuts that allow you to save and load two tinystates

idlechild:
People were kicking around the idea of allowing two tinystates within the full savestates build, and I think it should work.  We used to have some differences between fullstate and tinystates, but we've resolved enough differences to make it not that hard to add.

There's probably some things that could be adjusted.  Should there be more menu options?  Should the new read-only lock state take this into account?  I think I'm using the same lockstate for the tinystates; maybe that is fine?

Also included IFB's python changes, which looked fine, but I did notice an issue in defines.asm so I fixed that (actually I think there was an issue in defines.asm that got released but I don't think it would affect most people, and this might be ready to release soon so we can fix it then).


InsaneFirebat:
Its not supported in game over skip, crash handler, or the dev-only option to delete savestates.

idlechild:
Thanks for the feedback; I will look into those.

I'm also thinking about making separate read-only locks for the new save states.

idlechild:
Above issues addressed... this might be done?  Feedback from those who have tried it is positive.

PR #286, opened 2026-08-29
KPDR safeties first beta release

idlechild:
I should probably add a comment about the difference between an energy adjustment and a refill adjustment.  Energy adjustments play nice with other safeties, but sometimes you need the refill adjustment which does an overwrite.  Not sure where to put those comments though.

Also I expect there will be some adjustments before this is ready for release.  I'd also like to add phantoon-first safeties, but probably not going further than that.


idlechild:
I got a request for documentation and I think that's a good idea.  I'm going to try to add a new webpage, but not linked from the main page just yet.  That way I can just post a wall of text there and reference it from discord instead of posting wall of text to discord.

InsaneFirebat:
Is there a reason you don't use next-update?

InsaneFirebat:
I don't know how deep you are into this, but I'm willing to make a new preset movie if it'll help this out. I expect we'll be moving away from the original data and it might be good to get things updated before then.

idlechild:
I'm not putting in next update just because we might release next update before safeties.

Regarding preset movies question, I'm expecting preset categories made with safeties will not be compatible with movies.  Of course we can still have preset categories that don't use safeties and I expect a majority of our presets won't use them.  Hundo doesn't need safeties.  GT classic could have safeties but not sure it needs it.  I'm only planning to add safeties to Any% KPDR and Any% Phantoon First, and eventually KPDR No Drops just because I want to (not sure when I'm going to get to that).

PR #291, opened 2026-09-02
Fix potential crash when loading presets

InsaneFirebat:
I was experimenting with making demos load faster in the practice hack by copying `preset_start_gameplay`. I came across a crash after cutting out some things that were specific to presets. The hitbox color feature overwrote a stray RTL to expand the routine below it. `preset_start_gameplay` was calling the old address. The routine is supposed to start like this
```
load_common_target_colors:
    %ai16()
    LDX #$001E
  .first_loop
    LDA $9A81A0,X : STA $7EC3A0,X
    DEX #2 : BPL .first_loop
```
But it worked out to this on the first loop
```
    db $C2 ; the skipped byte that used to be RTL
load_common_target_colors:
    BMI $A2 ; crash if negative flag is set
    ASL $BF00,X
    LDY #$9A81
    STA $7EC3A0,X
    DEX #2
    BPL .first_loop
```
We don't notice the effects of this because X = $20, only one extra loop. The fix is replacing `JSL $82E139` with `JSL load_common_target_colors` in presets.asm

### DIFF
--- a/src/main.asm
+++ b/src/main.asm
@@ -17,7 +17,7 @@ lorom
 !VERSION_MAJOR = 2
 !VERSION_MINOR = 7
 !VERSION_BUILD = 11
-!VERSION_REV   = 0
+!VERSION_REV   = 1
 
 table ../resources/normal.tbl
 print ""
--- a/src/presets.asm
+++ b/src/presets.asm
@@ -758,7 +758,7 @@ endif
 
     JSL Randomize_Preset_Equipment
     JSL $90AC8D ; Update beam graphics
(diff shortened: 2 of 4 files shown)


idlechild:
Good find

This ended up being so minor I'm wondering if it's worth merging now (I'm not opposed to it)

InsaneFirebat:
It's on next-update now. I don't think it will crash but I didn't bother checking where the negative flag actually came from. It's passing through some PHP/PLP in the routine before it.

InsaneFirebat:
We should replace that changelog note if something more interesting is included.

